PR #51, opened 2014-07-19
Various simplification in trivial attributes.

Drup:
I think I'm going to wait before merging this one for the next major version of tyxml (along with some other breaking changes) in order not to break compat in each minor version of tyxml...

### DIFF
--- a/lib/html5_f.ml
+++ b/lib/html5_f.ml
@@ -277,7 +277,7 @@ struct
   let a_name = string_attrib "name"
 
   let a_autocomplete x =
-    user_attrib C.string_of_big_variant "autocomplete" x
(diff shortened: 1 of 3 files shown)


vasilisp:
JSOO looks unaffected. Eliom update in the works.

Drup:
@vasilisp I think it's best to wait for all the changes in tyxml to be done (that's only the start). Just use stable tyxml for now ...

vasilisp:
I had updated everything before reading your message (and we had broken compatibility with stable before). Also, if the other packages don't work with TyXML master, how are we going to find any issues?

Drup:
@vasilisp By testing before tyxml release ? :)
It's not going to be right now anyway.

PR #58, opened 2014-10-10
Rework wrapping

Drup:
This removes all the wrapped functor, simplify most signatures and remove the need for two arguments to use `Html5` signature (see [this](https://github.com/Drup/ocaml-markdown/commit/aab07fa115e2a72e78db09bfc7e50f8228ac678f)). 

This is done by adding a `W` module to `Xml` everywhere so that, given an `Xml` module (and, by extension, an `Html5` or `Svg` modules containing one, and all of them should), you can use the basic operations on the wrapped element.
The `'a wrap` and `'a list_wrap` are now just aliases. 

It also normalize the naming scheme. Every signatures (`Xml`, `Xml_wrap`, `Svg`, `Html5`) comes in two versions, the `T` version, which is the normal, fully general one and the `NoWrap` version, which comes with the forced equality `Xml.W = Xml_wrap.NoWrap`. 

It's breaking for everyone using the functorized interface but that's all.

In the future, we could provide a small functor taking a `W` and producing a minimal "standard library", I don't know if it's needed.

Basically, I did it painfully wrong the first time, and this fixes it. Sorry about it. :/
cc @darioteixeira, as he's our main user of tyxml's signatures.

@darioteixeira, the change needed after this for camlhighlight is implemented [here](https://github.com/Drup/camlhighlight/commit/5217053d4cadad99d986ef615a08f4cdc49b35ec). To accept an arbitrary wrapping, you replace the `NoWrap` by `T` and use the operations in `Html.Xml.W`.

### DIFF
--- a/lib/html5_f.ml
+++ b/lib/html5_f.ml
@@ -33,7 +33,7 @@
 open Html5_types
 
 module Make
-    (Xml : Xml_sigs.Wrapped)
+    (Xml : Xml_sigs.T)
     (Svg : Svg_sigs.T with module Xml := Xml)= struct
 
   module Xml = Xml
--- a/lib/html5_f.ml
+++ b/lib/html5_f.ml
@@ -36,9 +36,10 @@ module MakeWrapped
     (W : Xml_wrap.T)
     (Xml : Xml_sigs.Wrapped with type 'a wrap = 'a W.t
                              and type 'a list_wrap = 'a W.tlist)
-    (Svg : Svg_sigs.T with module Xml := Xml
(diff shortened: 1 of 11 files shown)


vasilisp:
@drup, this is fairly simple to rebase on top of current master (see branch `wrap_rebased`). Would you like to merge this before we release a new Eliom? If yes, this should happen soon. It is not a huge change, but we need synchronized updates in Eliom and jsoo.

Drup:
Small tyxml release done, let's merge all the things! \o/

PR #73, opened 2015-04-26
Fix svg element creation in Html5

pveber:
This PR follows from a suggestion by @Drup in an [jsoo issue](ocsigen/js_of_ocaml#276).

Description:
Use [Svg.svg] to create an svg embedded in a [Html5.elt] instead of [Xml.node].

### DIFF
--- a/lib/html5_f.ml
+++ b/lib/html5_f.ml
@@ -928,13 +928,8 @@ module MakeWrapped
 
   let form = star "form"
 
-  let svg ?(xmlns = "http://www.w3.org/2000/svg") ?(a = []) children =
-    let xmlns = string_attrib "xmlns" (W.return xmlns) in
-    let xmlns_xlink =
-      string_attrib "xmlns:xlink" (W.return "http://www.w3.org/1999/xlink")
-    in
-    let attribs = Svg.to_attrib xmlns :: Svg.to_attrib xmlns_xlink :: a in
-    Svg.toelt (Svg.svg ~a:attribs children)
+  let svg ?(a = []) children =
+    Svg.toelt (Svg.svg ~a children)
 
   let input = terminal "input"
 
--- a/lib/html5_f.ml
+++ b/lib/html5_f.ml
@@ -36,7 +36,8 @@ module MakeWrapped
     (W : Xml_wrap.T)
     (Xml : Xml_sigs.Wrapped with type 'a wrap = 'a W.t
                              and type 'a list_wrap = 'a W.tlist)
-    (Svg : Svg_sigs.T with module Xml := Xml)= struct
+    (Svg : Svg_sigs.T with module Xml := Xml
+                       and type 'a list_wrap = 'a W.tlist)= struct
 
   module Xml = Xml
 
@@ -928,13 +929,12 @@ module MakeWrapped
   let form = star "form"
 
   let svg ?(xmlns = "http://www.w3.org/2000/svg") ?(a = []) children =
-    let attribs =
-      string_attrib "xmlns" (W.return xmlns)
-      :: string_attrib "xmlns:xlink" (W.return "http://www.w3.org/1999/xlink")
-      :: Svg.to_xmlattribs a
+    let xmlns = string_attrib "xmlns" (W.return xmlns) in
+    let xmlns_xlink =
+      string_attrib "xmlns:xlink" (W.return "http://www.w3.org/1999/xlink")
     in
-    star ~a:(attribs)
-      "svg" (W.map Svg.toelt children)
+    let attribs = Svg.to_attrib xmlns :: Svg.to_attrib xmlns_xlink :: a in
+    Svg.toelt (Svg.svg ~a:attribs children)
 
   let input = terminal "input"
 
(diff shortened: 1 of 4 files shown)


Drup:
Svg.svg already define the xlink attribute, so you don't need to add it.

Do standalone svg document should have the xmlns attributes to ? if so, please add it to Svg.svg and fix Html5.svg, otherwise, just add the attributes a_xmlns in svg and use it in Html5. 

Better avoid custom redefinition of attributes.

pveber:
I'm not sure svg element *must* specify a namespace (standalone or not), but I'm pretty sure it doesn't hurt. In my last commit (998c568), I introduced an API change: it was possible with [Html5.svg] to specify an arbitrary namespace (but this was not the case with [Svg.svg]) using an optional argument called xmlns. I removed it (the only other possibility was to add one to [Svg.svg]) because I did not see any use for that. I think if we ever need to choose between several namespaces, a polymorphic variant argument to choose between a restricted set of possibilities would be better anyway.

pveber:
BTW, I can squash those two commits if you prefer a cleaner history.

lib/html5_f.ml
```diff
@@ -927,14 +928,8 @@ module MakeWrapped
...
-  let svg ?(xmlns = "http://www.w3.org/2000/svg") ?(a = []) children =
-    let attribs =
-      string_attrib "xmlns" (W.return xmlns)
-      :: string_attrib "xmlns:xlink" (W.return "http://www.w3.org/1999/xlink")
-      :: Svg.to_xmlattribs a
-    in
-    star ~a:(attribs)
-      "svg" (W.map Svg.toelt children)
+  let svg ?(a = []) children =
+    Svg.toelt (Svg.svg ~a children)
```
Drup: You can remove the `=[]` and the `a` as option (`?a`).

Drup:
is the list_wrap equality constraint still needed ?

Drup:
I think I would rather have the optional argument for xmlns everywhere rather than removing it. The transformation to a polymorphic sum type can wait some next contribution. ;)

pveber:
> is the list_wrap equality constraint still needed ?

I'm affraid it is, that or something else. If I remove the equality constraint on list_wrap, I obtain the following error:
```
File "lib/html5_f.ml", line 1:
Error: The implementation lib/html5_f.ml
       does not match the interface lib/html5_f.cmi:
       ...
       At position module Make(Xml)(Svg) : <here>
       Values do not match:
         val svg :
           ?a:[< Svg_types.svg_attr ] Svg.attrib list ->
           [< Svg_types.svg_content ] Svg.elt Svg.list_wrap -> Xml.elt
       is not included in
         val svg :
           ?a:[< Html5_types.svg_attrib ] Svg.attrib list ->
           [< Html5_types.svg_content ] Svg.elt list_wrap ->
           [> Html5_types.svg ] elt
       File "lib/html5_f.ml", line 931, characters 6-9: Actual declaration
```
which shows that in [Html5_f.Make], there is no way to know that the `'a list_wrap` is the same as `'a Svg.list_wrap`. Here is how I see the problem: the expected signature for `Html5.svg` requires a `Html5.list_wrap` of `Svg.elt`, but if I want to reuse `Svg.svg` I need to convert this `Html5.list_wrap` of `Svg.elt` to a `Svg.list_wrap` of `Svg.elt`.

One easy way to fix this (option 1) would be to change the expected signature of `Html5.svg` to:
```
val svg :
  ?a:[< Html5_types.svg_attrib ] Svg.attrib list ->
  [< Html5_types.svg_content ] Svg.elt Svg.list_wrap ->
  [> Html5_types.svg ] elt
```
If this is not acceptable, then it means we either have to say to the compiler `Svg.list_wrap` and `Html5.list_wrap` are the same (option 2, this is what I did), or provide a conversion function, for instance going back and forth the Xml representation (option 3). The trouble here is that the argument `Svg` to `Html5_f.Make` is typed as `Svg_sigs.T` where `list_wrap` is left abstract.

Personally, I'd favor option 1, which requires the minimum change and is conceptually in line with what this PR is about: putting SVG stuff in an HTML document :).

pveber:
> I think I would rather have the optional argument for xmlns everywhere rather than removing it. The > transformation to a polymorphic sum type can wait some next contribution. ;)

No problem to add it, but isn't this a useless complication for the API if it has no particular purpose? Or do you see some application to it? As a matter of comparison, there is no way to change the namespace for any other elements, in particular not `Html5.html`.

Drup:
Leave it how it is, I will get rid of all that when https://github.com/ocsigen/tyxml/pull/58 is merged anyway

> there is no way to change the namespace for any other elements, in particular not Html5.html.

Ok, point taken.

pveber:
Yeah, I agree #58 looks much cleaner anyway. IIUC, this reworking takes place in the `wrap` branch. If this branch is meant to overtake the current one, may I suggest to apply a similar change to it (because it currently bears the same problem that was corrected with this PR).

Drup:
Don't worry, it will be merged in due time. I'm actually waiting for the great renaming to break compat only once.

pveber:
OK, thanks!

2015-04-27 22:06 GMT+02:00 Gabriel Radanne <notifications@github.com>:

PR #76, opened 2015-08-03
Functorial interface updated to support shared reactive Eliom

vasilisp:
The `sharedreactcontent` branch provides the changes that we need in order to implement an `R` module for Eliom shared values.

The main points are:

* `Xml_wrap.T'` adds a type constructor `ft` to `Xml_wrap.T`. This is in order to accommodate
  shared functions for operating on shared values.
* A signature `Conv` contains instances of `ft` for all operations on wrapped elements that the
  functor performs. This is needed because we cannot simply inject server-side functions.
* Our use of shadow types becomes "customizable". The functor receives types `u` and `'a s` related
  by operations `shadow : u -> 'a s` and `unshadow : 'a s -> u` . `'a s` becomes the type of the resulting
  HTML elements. This is in order to achieve type equalities between different functor invocations (in
  Eliom) that would be hard to achieve otherwise.

The corresponding Eliom branch is also called `sharedreactcontent`.

Concerns? Suggestions? I know that the implementation is rough and incomplete, e.g., there is no SVG support yet. Hopefully an early pull request is a good format for discussion on this.

### DIFF
--- a/lib/html5_f.ml
+++ b/lib/html5_f.ml
@@ -32,11 +32,13 @@
 
 open Html5_types
 
-module Make
+module Make'_NoSVG
     (Xml : Xml_sigs.T)
-    (Svg : Svg_sigs.T with module Xml := Xml)= struct
+    (C : Html5_sigs.Conv with type ('a, 'b) ft = ('a, 'b) Xml.W.ft) =
+struct
 
   module Xml = Xml
+
   module W = Xml.W
 
   module Info = struct
@@ -79,83 +81,22 @@ module Make
 
   let user_attrib f name v = Xml.string_attrib name (W.fmap f v)
 
-  let bool_attrib = user_attrib string_of_bool
+  let bool_attrib = user_attrib C.string_of_bool
 
   (* space-separated *)
-  let length_to_string = function
-    | `Pixels p -> string_of_int p
-    | `Percent p -> (string_of_int p) ^ "%"
-
   let length_attrib name x =
-    user_attrib length_to_string name x
-
-  let multilength_to_string = function
-    | `Pixels p -> string_of_int p
-    | `Percent p -> (string_of_int p) ^ "%"
-    | `Relative 1 -> "*"
-    | `Relative i -> (string_of_int i) ^ "*"
-
-  let multilength_attrib name x =
-    user_attrib multilength_to_string name x
-
-  let multilength_to_string m =
-    String.concat ", " (List.map multilength_to_string m)
+    user_attrib
+      C.string_of_multilength
+      name (x :> [< Html5_types.multilength] W.t)
 
   let multilengths_attrib name x =
-    user_attrib multilength_to_string name x
+    user_attrib C.string_of_multilengths name x
 
-  let linktype_to_string =
-    function
-    | `Alternate -> "alternate"
-    | `Archives -> "archives"
-    | `Author -> "author"
-    | `Bookmark -> "bookmark"
-    | `External -> "external"
-    | `First -> "first"
-    | `Help -> "help"
-    | `Icon -> "icon"
-    | `Index -> "index"
-    | `Last -> "last"
-    | `License -> "license"
-    | `Next -> "next"
-    | `Nofollow -> "nofollow"
-    | `Noreferrer -> "noreferrer"
-    | `Pingback -> "pingback"
-    | `Prefetch -> "prefetch"
-    | `Prev -> "prev"
-    | `Search -> "search"
-    | `Stylesheet -> "stylesheet"
-    | `Sidebar -> "sidebar"
-    | `Tag -> "tag"
-    | `Up -> "up"
-    | `Other t -> t
+  let linktypes_attrib name x =
+    user_attrib C.string_of_linktypes name x
 
-  let linktypes_to_string l =
-    String.concat " " (List.map linktype_to_string l)
-
-  let linktypes_attrib name linktypes =
-    user_attrib linktypes_to_string name linktypes
-
-  let mediadesc_to_string =
-    function
-    | `All -> "all"
-    | `Aural -> "aural"
-    | `Braille -> "braille"
-    | `Embossed -> "embossed"
-    | `Handheld -> "handheld"
-    | `Print -> "print"
-    | `Projection -> "projection"
-    | `Screen -> "screen"
-    | `Speech -> "speech"
-    | `TTY -> "tty"
-    | `TV -> "tv"
-    | `Raw_mediadesc s -> s
-
-  let mediadescs_to_string mediadescs =
-    String.concat ", " (List.map mediadesc_to_string mediadescs)
-
-  let mediadesc_attrib name mediadescs =
-    user_attrib mediadescs_to_string name mediadescs
+  let mediadesc_attrib name x =
+    user_attrib C.string_of_mediadesc name x
 
   (* Core: *)
   let a_class = space_sep_attrib "class"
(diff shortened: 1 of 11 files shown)


lib/xml_wrap.ml
```diff
@@ -31,15 +31,49 @@ module type T = sig
...
+    ('a, 'b) ft -> 'a t -> 'b t
+
+  type 'a tlist
+  val nil : unit -> 'a tlist
+  val singleton : 'a t -> 'a tlist
+  val cons : 'a t -> 'a tlist -> 'a tlist
+  val append : 'a tlist -> 'a tlist -> 'a tlist
+  val map : ('a, 'b) ft -> 'a tlist -> 'b tlist
+end
+
```
Drup: No need to have two versions, almost nobody except eliom/jsoo is using T directly. Given the "most complete" version, you can trivially define the other with `T with type ('a, 'b) ft = 'a -> 'b` so it's not that much of an issue.

lib/xml_wrap.ml
```diff
@@ -31,15 +31,49 @@ module type T = sig
...
   type 'a tlist = 'a list
   external return : 'a -> 'a = "%identity"
   let fmap f :  'a t -> 'b t = f
 
+  let string_unoption = function
+    | None ->
+      ""
+    | Some s ->
+      s
+
```
Drup: Not sure why that's here. ;)

lib/xml_wrap.mli
```diff
@@ -30,5 +30,27 @@ module type T = sig
...
+  val map : ('a, 'b) ft -> 'a tlist -> 'b tlist
+
+end
+
 module NoWrap : T with type 'a t = 'a
                    and type 'a tlist = 'a list
+
+module NoWrap' : T' with type 'a t = 'a
+                     and type 'a tlist = 'a list
+                     and type ('a, 'b) ft = 'a -> 'b
```
Drup: the type of NoWrap' is more precise but included in the type of NoWrap, so no need for the two version, just define the most precise one (with the equalities).

Drup:
First things first: please rework based on https://github.com/ocsigen/tyxml/pull/58
I will merge #58 as soon as I can. It should makes things slightly simpler.

There are some things I don't like:
 - This forces the user to define some functions that should be defined by tyxml. You copied them to eliom, which is not really reasonable. Yes, we could provide a module with the functions predefined, but that's still a huge pile of complexity.
 - I'm not sure to understand what the whole shadow thing is for. It seems like a hack to avoid the current bug in eliom's typing on polymorphism, is that right ? It completely violate tyxml's guarantee, which I'm not fond of, to say the least.
The fact that you can't implement it without a Obj.magic (in eliom) may be a sign that somewhere, something is quite wrong. ;)

If I get things right, if we could define `pure : ('a -> 'b) -> ('a, 'b) ft`, it would be very easy. However, our main complicated case is `('a, 'b) ft = ('a -> 'b) shared_value`, which doesn't allow us to define `pure` on the server. Your current solution is to separate the set of pure functions in a module and make the user provide them.

If we had something of type `(P:Pairing) -> (H1: Html5_sigs) -> (H2: Html5_sigs) -> (H3: Html5_sigs with type 'a t = ('a H1.t, 'a H2.t) P.t)` (with P judiciously chosen :p), could we solve our problem ?

Let me think about it some more, to see if I can come up with a better solution, now that I understand the underlying issue.

lib/xml_wrap.ml
vasilisp: For no good reason. I was moving stuff around, and this was left behind :).

lib/xml_wrap.ml
vasilisp: I just tried to provide the exact interfaces that existed before, plus some more tedious ones for use by Eliom. Sure, we can remove the old ones if they are no longer needed.

vasilisp:
The shadow thing is a way to produce `'a Eliom_content_core.Html5.elt` instances outside of `Eliom_content_core`, without weakening the signature of `Eliom_content_core` with the equality `Xml.elt = 'a Html5.elt`.

We cannot put the shared reactive support inside `Eliom_content_core`, because shared values depend on a whole set of modules (`Eliom_service` and friends), which in turn depend on `Eliom_content_core`.  But we do need to maintain the equality on the types of elements; if our reactive module does not produce the standard kind of elements, then it becomes unusable.

Nothing stops TyXML from providing the same guarantees as before; note that the signature of `MakeWrapped` has not changed (apart from an accidental elimination of the variance annotations, which is not a big issue). For the lower-level interfaces, restricting the signature can happen outside the functor.

The `%identity` stuff is just a workaround for Eliom's lack of polymorphism, and not related to the core issue :).

vasilisp:
Indeed, the inability to define `pure` is the core issue here. If we cannot define `pure`, then there is no way of internally producing shared functions out of the little functions in `Html5_f`, so we cannot operate on shared reactive signals. Also, if we cannot define `pure`, then a nice functorial abstraction is not possible, and the little functions need to be tediously implemented.

Drup:
> The shadow thing is a way to produce 'a Eliom_content_core.Html5.elt instances outside of Eliom_content_core, without weakening the signature of Eliom_content_core with the equality Xml.elt = 'a Html5.elt.

Why ? As long as it's not exported in `Eliom_content`, it's not an issue. I think you should try that instead.

In eliom, you do `pcdata x |> C.unshadow |> C.shadow` which is equivalent to throwing away tyxml's garantee, so you *do* weaken something at some point.

vasilisp:
> Why ? As long as it's not exported in `Eliom_content`, it's not an issue. I think you should try that instead.

`Eliom_content_core` is used internally. At the very least, we would be enabling a class of bugs in Eliom itself.

> In eliom, you do `pcdata x |> C.unshadow |> C.shadow` which is equivalent to throwing away tyxml's garantee, so you do weaken something at some point.

That's not much different from the `tot` / `toelt` conversions that TyXML already provides (and that Eliom uses in many places). We do allow conversions between type and untyped elements, and for good reason. That's more controlled and safe than a type equality between typed and untyped elements.

Also, note that the conversion you point to is in an implementation context, and it does not prevent us from producing a well-behaved `R` module in the end.

Drup:
If it's equivalent to `tot/toelt`, why do you need it ?

vasilisp:
Because I need to re-use a previously existing kind of untyped XML elements and a previously existing kind of typed HTML elements (from `Eliom_content_core`). The convertibility between the two has to be established outside the functor, so that we don't end up with a seemingly different kind of HTML elements.

It is "equivalent" with `tot` / `toelt` only with respect to converting back and forth between typed and untyped elements.

Think of this a feature for producing the same old elements with a different module (that implements shared reactive updates in our case). I could call it "embedding" (into the old universe of typed HTML elements), probably calling it "shadowing" was unfortunate and leads to confusion.

vasilisp:
I withdraw the "embedding" part of the request :). I think I can achieve what I want via a functor in `Eliom_content_core`. This had not occurred to me previously.

The rest remains relevant, unfortunately.

Drup:
Small tyxml release done, let's merge all the things! \o/

balat:
If we see no other solutions for avoiding all the functions I suggest we merge this (as other pull request depend on this in js_of_ocaml and eliom). @Drup ?

lib/html5_sigs.mli
```diff
@@ -1190,6 +1190,46 @@ end
 
 module type NoWrap = T with module Xml.W = Xml_wrap.NoWrap
 
+module type Conv = sig
+
+  type (-'a, 'b) ft
```
Drup: Isn't 'b covariant ? Might as well mark it too.

lib/html5_f.ml
```diff
@@ -32,11 +32,16 @@
 
 open Html5_types
 
-module Make
+module Make'
```
Drup: Please use something more descriptive than `Make'`. Same for `Conv`.

Maybe `Constants` and `Make_with_constants` ?

Drup:
@darioteixeira : How much more painful does it make your usage of tyxml ?

Drup:
@vasilisp Could you update the documentation to explain both th rational behind this additional argument and how to use it ?

vasilisp:
@Drup , I will, but basically you don't need to use it unless you are Eliom :). We still provide the old functorial interface (that simply applies the internal `Conv`). The only difference for non-Eliom users should be defining `type (-'a, 'b) ft = 'a -> 'b`.

darioteixeira:
> @darioteixeira : How much more painful does it make your usage of tyxml ?

Thanks for the heads-up, @drup. If, as @vasilisp said, the only difference for non-Eliom users should be defining `type (-'a, 'b) ft = 'a -> 'b`, then it won't be too much of a burden.

Drup:
@vasilisp That's if you use the functorial interface to define a new html module based on some custom Xml implementation. @darioteixeira takes an arbitrary Html module and defines things with it, which is a slightly different way to use tyxml's signatures.

Drup:
After re-reading the code after the various patches, and in particular now that the shadow thing is gone, I'm happy with the patch and I convinced myself it was a decent solution. :) 

Just merge once the various naming/documenting tibits are fixed.

lib/html5_f.ml
vasilisp: @drup, how about `Wrapped_functions` and `Make_with_wrapped_functions`? Constants is at least vague.

Drup:
Good for me.

Drup:
LGTM!

PR #78, opened 2015-10-19
Use Xml_print.string_of_number to print floats.

Drup:
Luckily, svg, html5 and javascript's norms seem to agree on how numbers should be printed!

Our unit printing is slightly mess, we should clean that up at some point ...

cc @mfp


vasilisp:
The patch looks good to me.

PR #79, opened 2015-11-21
Add <main> tag

slegrand45:
Hi,
This is a patch proposal to add the `<main>` tag. I am really not sure if the `core_flow5` type is the best place for this tag. The HTML5 spec specifies two restrictions: `<main>` must not be a descendent of an `<article>`, `<aside>`, `<footer>`, `<header>`, or `<nav>` element and authors must not include more than one `<main>` element in a document. But the living HTML standard removes these two restrictions.

ref. :  https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main and https://html.spec.whatwg.org/multipage/semantics.html#the-main-element

### DIFF
--- a/lib/html5_f.ml
+++ b/lib/html5_f.ml
@@ -661,6 +661,8 @@ struct
 
   let aside = star "aside"
 
+  let main = star "main"
+
   let video_audio name ?src ?srcs ?(a = []) elts =
     let a =
       match src with
--- a/lib/html5_sigs.mli
+++ b/lib/html5_sigs.mli
@@ -603,6 +603,9 @@ module type T = sig
   val aside :
     ([< | aside_attrib], [< | aside_content_fun], [> | aside]) star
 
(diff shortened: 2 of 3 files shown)


Drup:
I agree with you, the living standard takes precedence. There are apparently discussions to reintroduce the uniqueness constraint, but we can't enforce that, so it doesn't matter. 

Could you please add the whatwg link inside the commit message ? I will merge when that's done.

slegrand45:
The whatwg link is now inside the commit message.

Drup:
I just realized you missed a good amount of things html5_types. You should also add `Main to all the `flow5_without_*` types.

Drup:
@slegrand45 There is no notification when a commit is pushed on a PR, you need to write a message. ;)

slegrand45:
Ok, sorry :) Thanks for the patch review.

PR #80, opened 2016-02-06
Add srcset attribute in <img>

slegrand45:
https://www.w3.org/html/wg/drafts/html/master/semantics.html#attr-img-srcset

See also for instance: 
https://developer.mozilla.org/en-US/Learn/HTML/Howto/Add_responsive_image_to_a_webpage
https://bocoup.com/weblog/smaller-faster-websites
https://www.smashingmagazine.com/2014/05/responsive-images-done-right-guide-picture-srcset/

### DIFF
--- a/lib/html5_f.ml
+++ b/lib/html5_f.ml
@@ -1054,12 +1054,12 @@ module Wrapped_functions = struct
 
   let string_of_srcset l =
     let f = function
-    | `Img_cand_url url ->
+    | `Url url ->
       Xml.string_of_uri url
-    | `Img_cand_width (url, v) ->
+    | `Url_width (url, v) ->
       (Xml.string_of_uri url) ^ " " ^ (string_of_number v) ^ "w"
-    | `Img_cand_pixel (url, v) ->
-      (Xml.string_of_uri url) ^ " " ^ (string_of_float v) ^ "x"
+    | `Url_pixel (url, v) ->
+      (Xml.string_of_uri url) ^ " " ^ (Xml_print.string_of_number v) ^ "x"
     in
     String.concat ", " (List.map f l)
 
--- a/lib/html5_f.ml
+++ b/lib/html5_f.ml
@@ -99,6 +99,9 @@ struct
   let mediadesc_attrib name x =
     user_attrib C.string_of_mediadesc name x
 
+  let srcset_attrib name x =
+    user_attrib C.string_of_srcset name x
+
   (* Core: *)
   let a_class = space_sep_attrib "class"
 
@@ -393,6 +396,8 @@ struct
 
   (*let a_srcdoc*)
   let a_srclang = string_attrib "xml:lang"
+
+  let a_srcset = srcset_attrib "srcset"
 
   let a_start = int_attrib "start"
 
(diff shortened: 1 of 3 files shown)


vasilisp:
The `` `Img_cand_width `` and `` `Img_cand_pixel `` names are a bit confusing. What about something closer to the terms used in the W3C doc, e.g., `` `Img_candidate_ratio`` and `` `Img_candidate_viewport``? Looks good other than that.

Drup:
`string_of_float` is incorrect for most purposes in html. There is a correct implementation in `Xml_print.string_of_number` ... but you can't use that directly. Look at how it's done in `Svg_f`, you need to add a new function to `Html_sigs.Wrapped_functions` and provides the good default implementation.

@vasilisp I read "width descriptor" and "pixel density descriptor". Where do you read ratio and viewport ? 

As opposed to @vasilisp, and since those are not phantom, just use shorten names : ``` `Url | `Url_with | `Url_pixel ```.

slegrand45:
@Drup i'm a bit confused when you say that i can't directly use `Xml_print.string_of_number` because there is already a such use in `string_of_step` function: https://github.com/ocsigen/tyxml/blob/master/lib/html5_f.ml#L1024-L1026 (?).

Drup:
Hum. @vasilisp why is that not in the wrapped functions ?

vasilisp:
@Drup for the names, I was looking at the description of the [selection methods](https://www.w3.org/html/wg/drafts/html/master/semantics.html#introduction-3:device-pixel-ratio-2). But you are right, I hadn't seen the names of the descriptors, which are indeed "width" and "pixel density".

`string_of_step` is in `Wrapped_functions`. Eliom uses client and server instances of it to build an `ft` and things work out; it suffices that the indirectly used `Xml_print.string_of_number` is available on both sides, and it doesn't have to appear directly in `Wrapped_functions`. `string_of_srcset` can similarly use `Xml_print.string_of_number`.

As a bit of context, it was easier to get the Eliom types working with more complex functions like these that operate over options (for instance) rather than with higher-order combinators. Hence there is no `Xml_print.string_of_number`.

Drup:
Ok, then you can use `string_of_number` directly, @slegrand45

slegrand45:
Patch updated yesterday. But the automatic checks seem stuck?

lib/html5_f.ml
```diff
@@ -1047,6 +1052,17 @@ module Wrapped_functions = struct
...
     String.concat " " (List.map string_of_linktype l)
 
+  let string_of_srcset l =
+    let f = function
+    | `Url url ->
+      Xml.string_of_uri url
+    | `Url_width (url, v) ->
+      (Xml.string_of_uri url) ^ " " ^ (string_of_number v) ^ "w"
+    | `Url_pixel (url, v) ->
+      (Xml.string_of_uri url) ^ " " ^ (Xml_print.string_of_number v) ^ "x"
```
Drup: Either String.concat or Printf, avoid series of `^`.

slegrand45:
The patch now uses Printf.sprintf instead of ^

Drup:
@vasilisp Do you know what's happening with jenkins ?

Issue #81, opened 2016-02-16
The functor version is currently broken


vouillon:
I get this error:
```
Error: Error while linking /home/jerome/.opam/4.02.3/lib/tyxml/tyxml_f.cma(Html5_f):
Reference to undefined global `Xml'
```
I believe this is because we now to refer to `Xml.string_of_uri` in module `Wrapped_functions` in `html5_f.ml`.

I don't know how to fix that.

vasilisp:
I don't understand the underlying issue, but a quick workaround is to remove the `Xml.string_of_uri` calls (branch `workaround-81`); `Xml.uri = string` and the conversion is the identity function.

vasilisp:
@Drup, maybe there should be a `string_of_uri` in `Xml_print` instead of Xml? Is the current `string_of_uri` a placeholder for something that encodes special characters?

vouillon:
The `Xml` module is not included in the functor version of the library...

Drup:
Fixed by https://github.com/ocsigen/tyxml/pull/82

PR #82, opened 2016-02-16
Work around issue #81

vasilisp:
This is not a long-term solution, because it exploits the accidental type equality between URIs and strings. It suffices to keep Eliom et al. happy for now.


Drup:
Good enough for now, we will need to fix that at some point (and to figure out something for constants functions, in general).

Issue #72, opened 2015-04-03
Consistent naming scheme for html and svg elements.


Drup:
Consistency is not W3C's strong point, but it's made worse in tyxml. The main pain point is dashes, since they are not valid in ocaml's identifiers.

So a more formal version of what we are currently loosely applying:
- We *always* keep the name in the specification if it's a valid ocaml identifier, be it snake_case, nocase, CamlCase or whatever, even if it's internally inconsistent (I'm looking at you, `fontFace`.)
- Dashes are transformed in underscores systematically.
- Underscores are kept.
- Any other characters that could ever appear are transformed in an underscore too. I don't think I ever saw one, but better be sure.
- In case of conflict between attributes of incompatible types. We put the element that it's disambiguated for between underscores, after the `a`.

  Example: the attribute `max` for `input` is transformed in `a_input_max`.
- Specific transformation rules: 
  - When transforming to an element, the first letter become a lowercase.
  - When transforming to a polymorphic variant, the first letter is capitalized. This is not mandatory (since polymorphic variant can start with a lowercase), but it's acceptable and is already applied almost consistently, so better keep it.
  - When transforming to an attribute, no letter is changed (it's ok, it starts with `a_`).

These should also keep the breaking rather minimal. Remarks ?
cc @dsheets since you might have an opinion. :)

darioteixeira:
Seems reasonable.  Though W3C's "conventions" are inconsistent, I agree it is better to have a fairly verbatim mapping between them and Tyxml's names, as I reckon most developers rely on resources such as [MDN](https://developer.mozilla.org/), which of course use W3C's naming.

Drup:
Apparently, some svg elements have ':' in the name. so that would become a '_' in this scheme.

Drup:
I pushed a bunch of functions and a little tool called `autoname` that does the transformation automatically. 

The only things remaining to do is to transform the whole codebase .... *sob*.

Drup:
The situation is much better now, but not yet enforced by the ppx. We will get there eventually.

PR #83, opened 2016-02-20
Add sizes attribute in <img> tag

slegrand45:
https://www.w3.org/html/wg/drafts/html/master/semantics.html#attr-img-sizes
https://www.smashingmagazine.com/2014/05/responsive-images-done-right-guide-picture-srcset/

As this attribute can be quite complex (for instance: `(min-width: 36em) calc(.333 * (100vw - 12em)), 100vw`), a string list seems a good realistic choice.

### DIFF
--- a/lib/html5_f.ml
+++ b/lib/html5_f.ml
@@ -102,6 +102,8 @@ struct
   let srcset_attrib name x =
     user_attrib C.string_of_srcset name x
 
+  let src_sizes_attrib = comma_sep_attrib "sizes"
+
   (* Core: *)
   let a_class = space_sep_attrib "class"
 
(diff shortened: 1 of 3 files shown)


Drup:
Is the [w3c's grammar](http://w3c.github.io/html/semantics.html#valid-source-size-list) buggy ? In theory, I would propose to use a list of pair (media query * length) plus an argument length, but those are css length, we don't type those (and it would be complex, as you pointed out).

The naming doesn't follow https://github.com/ocsigen/tyxml/issues/72. Please change it to `a_img_sizes`

slegrand45:
Renaming done.

lib/html5_f.ml
```diff
@@ -102,6 +102,8 @@ struct
   let srcset_attrib name x =
     user_attrib C.string_of_srcset name x
 
+  let img_sizes_attrib = comma_sep_attrib "sizes"
+
```
Drup: What is this for ?

slegrand45:
Indeed, img_sizes_attrib is completely useless (a residue of previous attempts). Removed.

Drup:
Thanks !

PR #84, opened 2016-02-21
Fix inconsistent naming and typos.

aantron:
Per #72.

This doesn't block the ppx – I have `a_sandbox` special-cased for the time being.


aantron:
This still doesn't block the ppx, except that it will be difficult to test it with SVG without writing a lot of special cases/reflection annotations to translate the presently wrong attribute names. There is no point in doing that, since (I hope) the attributes will be renamed in the future anyway.

Drup:
Let's start merging things for 4.0 !

aantron:
:+1:

PR #87, opened 2016-02-24
Add basic, statically generated website example

fxfactorial:
Following up on IRC request.

### DIFF
--- a/examples/basic_website/.merlin
+++ b/examples/basic_website/.merlin
@@ -1,0 +1,1 @@
+PKG tyxml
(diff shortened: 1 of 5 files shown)


Drup:
Thanks !

Drup:
@fxfactorial Do you agree to put the example under public domain (or http://unlicense.org/, whichever you prefer) ?

I don't want it in LGPL, since it's intended to be copied.

fxfactorial:
@Drup whatever is best for tyxml is fine for me

PR #88, opened 2016-02-25
Replace some option-like variants by options.

aantron:
This is consistent with how `Html5_sigs.T.a_step` is represented, and makes the ppx ever-so-slightly less tedious to program. Curious to hear about the impact on users.

**Edit**: also replacing type of `a_sizes`. I suppose we could also go the other way, keep these as they are, but change `a_step` to a variant.

### DIFF
--- a/lib/html5_f.ml
+++ b/lib/html5_f.ml
@@ -1038,10 +1038,10 @@ module Wrapped_functions = struct
     | None -> "any"
 
   let string_of_sizes = function
-    | `Sizes l ->
+    | Some l ->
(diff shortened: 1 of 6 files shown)


Drup:
In general, I'm of the opinion we should move towards native ocaml types when it makes sense (such as the recent removal of On/Off). The reason I didn't changed that one is because `None` and `` `Auto` don't really mean the same thing, so I'm not sure how easy to understand it is.

I'm rather in favor, though, so if nobody complains, I will merge.

PR #92, opened 2016-03-21
Add testing infrastructure.

Drup:
The build system part is quite repugnant, but surprisingly, it seems to work.

@gasche: I had trouble to create a rule that doesn't produce anything. Is this use of stamp correct ?
Also, if anyone has a better solution to avoid all the load/requires at the top of the files (and better integrate with oasis), that would be nice.

We only need to add tests now ...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ setup.exe
 setup-dev.exe
 *.byte
 *.native
+_tests
--- a/.jenkins.sh
+++ b/.jenkins.sh
@@ -1,7 +1,7 @@
 opam pin add --no-action tyxml .
 opam install camlp4
-opam install --deps-only tyxml
-opam install --verbose tyxml
+opam install -t --deps-only tyxml
+opam install -t --verbose tyxml
 
 do_build_doc () {
   make wikidoc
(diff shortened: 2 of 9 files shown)


gasche:
Seems ok (I wouldn't know without testing really, but I would assume you did), exception I would call the file `.diff.stamp` or just `.stamp` instead of `.diff`: this suggests it contains a diff output, while it does not.

If I may say, this is rather the old-school approach to testing typing error messages. The cool kids [are using 'expect' now](https://github.com/ocaml/ocaml/pull/457).

Drup:
@gasche: I just tried the "cool kids" method. It means I have to include a file that is 1) a bit complicated 2) extremely tied to compiler-libs, which means it's going to break at every caml release. The build script is exactly as complicated as before and the end result is not that much of a benefit (inlining error messages when the error message is more than a screen longs is not really fabulous).

So, I will keep the old school method.

Drup:
@gasche I renamed *.diff to *.stamp, thanks!

Drup:
@gasche Hum, it doesn't exactly work in practice, it doesn't always rebuild the target. How do I make a real phony target in ocamlbuild ?

gasche:
Shouldn't you rather add a dependency on `lib/tyxml.cma`?

Issue #22, opened 2014-01-22
Xml & Svg : namespace conflict  


hhugo:
`Xml` & `Svg` can easily collide with other libraries.
It would be better to have them packed.

sagotch:
While waiting for a fix, is there any workaround for using another lib with a xml module?

```
ocsigenserver: ocsigen:main: Fatal - While loading /[...]/.opam/4.02.3/lib/xml-light/xml-light.cma: interface mismatch on Xml
```

Drup:
Fixed in master.

PR #94, opened 2016-03-22
Split implementation and make uri abstract again

Drup:
I tried several approach, but was more or less forced to do it that way.
@vasilisp : What's your opinion ? I decided to parametrized the wrapped function signature over the whole Xml, instead of a selected subset of it, since it looked more future-proof.

The spliting of directory is to enforce the separation.
A side effect of this change is to make the _f _sigs _types modules visible when just using `tyxml` (and not `.functor`). I used to [feel strongly about it](https://github.com/ocsigen/tyxml/issues/22#issuecomment-33071670), but it doesn't really matter ...


vasilisp:
I get your points, but unfortunately things are very messy with `Eliom_shared_content` in the picture.

We currently use client and server copies of `Wrapped_functions` to produce a version that has `type (-'a, 'b) ft = unit -> ('a -> 'b) shared_value`. The new interface would necessitate calling the functor on the client and on the server, and then coupling the functions. In idealized Eliom, we would do something like this:

```ocaml
module Wrapped_functions (Xml : Xml_sigs.T with type ('a,'b) W.ft = 'a -> 'b) = struct

  module Raw = Html5_f.Wrapped_functions(Xml)

  module%client Raw = Html5_f.Wrapped_functions(~%Xml)

  let string_of_bool () = [%shared Raw.string_of_bool ]

  (* ... *)

end
```

But of course this hits the fundamental limitations of the Eliom language. We can't produce an `~%Xml` that matches the `Xml` argument to our functor.

In practice, we will only use this with a fixed `Xml` module, so I can imagine doing ugly hacks that provide a pseudo-functorial interface to make the compiler happy. However, if we provide an interface with a unique use case, then this use case shouldn't need to do things like this...

vasilisp:
@Drup, an `Xml` under `Svg_sigs.Wrapped_functions` would be nice (for consistency with `Html`).

Drup:
> @Drup, an Xml under Svg_sigs.Wrapped_functions would be nice (for consistency with Html).

Done in https://github.com/ocsigen/tyxml/pull/124

PR #105, opened 2016-04-02
Restore compatibility with 4.03

Drup:
We have lot's of deprecation warnings, but it's not possible to fix them
while still keeping 4.02 compatibility.


Drup:
(This one will only be merged when ppx_tools is released)

Drup:
This is a tad annoying, due to change in the typechecker error messages, the html_fail tests only works on 4.03 or 4.02.

PR #107, opened 2016-04-02
Various improvements to the ppx:

Drup:
I removed the tyxml quotation. The namespace is now always provided by the quotation.

- Antiquotations inside lists (that is, star elements) now should be lists.
  Antiquotations inside unary elements can still be single elements
- The ppx will return a single element if there is only one element in the quotation.
- pcdata and svg-inside-html handling is probably a bit better now.
- Some judiciously placed type constraint allow better type errors for quotations.

@aantron: I think I broke `adjust_location` in one of my patches ...

### DIFF
--- a/_oasis
+++ b/_oasis
@@ -225,6 +225,20 @@ Test mini_website
   TestTools: mini_website
   Run$: flag(tests)
 
+Executable mini_website_ppx
+  Install: false
+  Build$: flag(tests) && flag(tests)
+  Path: examples/mini_website_ppx
+  MainIs: minihtml.ml
+  BuildDepends: tyxml
+  CompiledObject: best
+
+Test mini_website_ppx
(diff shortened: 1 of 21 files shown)


Drup:
@aantron In the end, I did the namespace thing in a slightly different way: Since we always have the parent namespace, I can just propagate it in a top-bottom fashion (it was done bottom-top previously). This allows to remove `qualify_children`, which is good.

aantron:
Yep, this sounds good.

That was the main reason I didn't fix qualifying the top-level `pcdata` elements in the original PR. I wanted to discuss first what quotation we would be using and how the namespace would be selected/discovered – since the top-down approach makes this a non-issue.

Drup:
> since the top-down approach makes this a non-issue.

Yes, It just took me a week to understand it. :p

Issue #90, opened 2016-03-08
Comments, RAWTEXT, script data, etc., should not be escaped


aantron:
See the [specification](https://www.w3.org/TR/html5/syntax.html#serializing-html-fragments) for serialization.

However, this program

```ocaml
open Html5.M
open Html5.P

let () =
  print_list print_string
    [tot (Xml.comment "foo&");
     script (pcdata "foo&");
     style [pcdata "foo&"]]
```

produces this output:

```html
<!--foo&amp;--><script>foo&amp;</script><style>foo&amp;</style>
```

In particular, this seems to prevent comments such as:

```html
<!--[if IE 8]> <html class="no-js lt-ie9" lang="en"> <![endif]-->
```

being generated by expressions such as

```ocaml
tot (Xml.comment "[if IE 8]> <html class=\"no-js lt-ie9\" lang=\"en\"> <![endif]")
```

(credit to @sgrove for a query).

In the case of comments, if the user includes `-->` as text, we obviously have a problem if we "fix" this issue. Perhaps we can escape only those. Similarly, in RAWTEXT of a `style` tag, we may have to handle `</style>`.

Rather than fixing this immediately in TyXML, I propose that before TyXML 4.0,

- I split up Markup.ml into subpackages, or aliased modules, one of which will include the HTML writer, but exclude any other writers or parsers,
- I add a polyglot output mode to Markup.ml,
- Markup.ml (IIRC) already properly *doesn't* escape the above fragments, but we decide what to do about `-->` in comments, etc., and do it, or support doing it, in Markup.ml,
- we use Markup.ml's HTML writer to do the HTML writing.

This will leave us with only one place at which to maintain conformance with HTML5 and XHTML. It will add Markup.ml as a run-time dependency to projects using TyXML (as opposed to only build-time with the PPX). The split of Markup.ml into packages or aliased submodules should prevent the linking of unnecessary code into users' programs, such as the enormous HTML5 parser.

Drup:
> It will add Markup.ml as a run-time dependency to projects using TyXML

This may not be true, Tyxml is currently split between the functor part and the concrete implementation. I have no problem adding dependencies to the concrete implementation (especially a printer) since it's not used inside eliom and js_of_ocaml.

Drup:
This is now partially solved, for comments. Export to markup printer is still desirable but should be addressed in https://github.com/ocsigen/tyxml/issues/100

Unescaped data (for script/style) can be introduced with `Unsafe.data`.

[...]

PR #110, opened 2016-04-02
Do not escape comments

Drup:
See https://github.com/ocsigen/tyxml/issues/90
This is only a simplistic fix (and it adds re as a dep of the functor part, which is annoying).

### DIFF
--- a/lib/xml_print.ml
+++ b/lib/xml_print.ml
@@ -48,16 +48,39 @@ let encode_unsafe_char_and_at s =
   Buffer.contents b
 
 let compose_decl ?(version = "1.0") ?(encoding = "UTF-8") () =
-  "<?xml version=\"" ^ version ^ "\" encoding=\"" ^ encoding ^ "\"?>\n"
+  Format.sprintf
+    {|<?xml version="%s" encoding="%s"?>\n|}
+    version encoding
 
 let compose_doctype dt args =
-  "<!DOCTYPE " ^ dt
-  ^ (if args = []
-     then ""
-     else
-       " PUBLIC " ^
-       String.concat " " (List.map (fun a -> "\"" ^ a ^ "\"") args)) ^ ">"
+  let pp_args fmt = function
+    | [] -> ()
+    | l ->
+      Format.fprintf fmt " PUBLIC %a"
+        (Format.pp_print_list ~pp_sep:Format.pp_print_space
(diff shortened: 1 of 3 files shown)


lib/xml_print.ml
```diff
@@ -48,16 +48,27 @@ let encode_unsafe_char_and_at s =
...
+        l
+  in
+  Format.asprintf
+    "<!DOCTYPE %s%a>"
+    dt
+    pp_args args
 
+let re_end_comment = Re.(compile @@ str "-->")
+let escape_comment =
+  Re.replace_string ~all:true re_end_comment ~by:"--&gt;"
```
aantron: In HTML (and IIRC XML), even `--` is not allowed in comments, so it may be more sensible to replace `-+` (as a regexp) with `-`. Also, the replacement `&gt;` doesn't have any special meaning in comments.

aantron:
Actually, in HTML5, `<!-->` is unfortunately a comment as well, so you need to avoid payloads beginning `>`. My commit comment is not sufficient to deal with this. Anyway, HTML is a giant mess and I need to take a break, but if you ignore this issue for a few hours, I will come back with some recommendation and a list of what things to watch out for, based on the comment parsing rules in the HTML spec.

aantron:
Ok, here are all the cases for a comment ending "prematurely":

1. the comment data starts with `>`,
2. the comment data starts with `->`,
3. the comment data contains '-->',
4. the comment data contains '--!>'.

Only (3) represents a valid comment terminator, but recovering parsers (such as in browsers) are required to end the comment in all of these situations.

(4) is a bit irregular: comment data can start with `!>` or `-!>` without triggering a premature comment end.

Probably, replacing `>` with `&gt;` in each of these cases is the best approach. It applies systematically, and is pretty easy to reason about since it `&gt;` doesn't contain `-`, `!`, `>`.

Drup:
@aantron Thanks a lot, I modified the code to handle all the cases.

aantron:
:+1: LGTM

aantron:
I should qualify that – I looked at the comments. The XML declaration and DOCTYPE output code does look a bit sketchy, depending on where the arguments can come from.

Drup:
> The XML declaration and DOCTYPE output code does look a bit sketchy, depending on where the arguments can come from.

I doubt people will have user input ending up there.

PR #128, opened 2016-04-18
Type safe templating (with mustache)

Drup:
This is WIP.

```ocaml
open Tyxml

let%template f = {|
Hello {{name}}
You have just won {{value}} dollars!
{{#repo}}
  <b>{{foo}}</b> {{name}}
{{/repo}}
{{^repo}}
  No repos :(
{{/repo}}
|}

val f :
  value:bytes ->
  repo:< foo : bytes; .. > list ->
  name:bytes -> [> `B | `PCDATA ] Tyxml_html.elt list
```

Current issues:
- Locations are completely lost (because of the mustache implementation)
- Nesting is untested
- Can't specify a name relative to the current context. Names are always resolved from the top context.


aantron:
I think this is (potentially) a great idea, but I don't think TyXML should offer two PPXs as in the attached commits. Ideally, there would be "one great TyXML PPX" that offers this and whatever other features. Otherwise, we are likely to confuse new users and make them anxious about going down the wrong path with their code.

Drup:
If people want this to happen, please voice your opinion on the choice of the templating language in https://discuss.ocaml.org/t/choice-of-template-language-for-tyxml/1863 .

PR #130, opened 2016-04-19
Implement let%html constructs

Drup:
This works with functions and multiple bindings. In particular, something like that works:

```ocaml
let%html make_content content = {|
  <div class="content">|} content {|</div>
|}

let%html rec f = function
  | [] -> ""
  | h :: t -> "<li>" [h] "</li>" (f t)
val f :
  [< Html_types.li_content_fun ] Tyxml_html.elt list ->
  [> Html_types.li ] Tyxml_html.elt list
```

(And now that I've written out the second example, I'm not sure it's a great idea anymore :D)

### DIFF
--- a/doc/manual-wiki/ppx.wiki
+++ b/doc/manual-wiki/ppx.wiki
@@ -59,6 +59,20 @@ let my_div = [%html "<div id="my_id"></div>"]
 val my_div : [> Html_types.div ] Html.elt
 >>
 
+==@@id="let"@@ Let notation
+
+It is also possible to use the ppx with the ##let## notation:
+<<code language="ocaml"|
+let%html content = {|<div id="content">some content</div>|} ;;
+val content : [> Html_types.div ] Html.elt
+>>
+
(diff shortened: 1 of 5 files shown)


aantron:
What's the advantage of this over `[%html ...]`? It's slightly more concise to type, or there are some new possibilities?

Drup:
It's only more concise, the expressive power is exactly the same.

aantron:
Ok. The second example *is* neat.

aantron:
Still suggest `match%html` to go with this (not sure if you're already on it), in case someone ever refactors a `function`. And I'm assuming there will be `%svg` too, for consistency? Or it's not worth it?

Drup:
`let%svg` is already there, along with `let%html.M` and `let%tyxml.html`. It's the same codepath than regular quotation.

I really don't like `match%html`, it implies things about matching that are just not true.

aantron:
> it implies things about matching that are just not true.

What does this refer to?

Drup:
if you see in a code a construction
```
match%html bla with
| ...
```

What do you think first, that it emits html or that it matches html ?

aantron:
Ok, wanted to make sure.

PR #132, opened 2016-04-19
Fix capitalization in Html5_types.mediadesc_token

aantron:
Without this change, the PPX generates `` `Raw_mediadesc "tv"`` for `media=tv`, and similarly for `media=tty`.

Another way to solve this is to do a case-insensitive search for a matching variand. That slightly decreases the consistency of the API, but would allow intuitive labels `` `TV`` and `` `TTY`` to work "correctly" with the PPX. Let me know if you prefer that.

### DIFF
--- a/lib/html_f.ml
+++ b/lib/html_f.ml
@@ -875,8 +875,8 @@ struct
     | `Projection -> "projection"
     | `Screen -> "screen"
     | `Speech -> "speech"
-    | `TTY -> "tty"
-    | `TV -> "tv"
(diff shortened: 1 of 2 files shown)


Drup:
It's better to do it that way, but I think we should have case insensitive matching anyway. Iirc, attributes are case insensitive in the spec.

aantron:
Attribute names are case-insensitive, not sure about attribute values in general. The parser preserves their case, though for these specific attribute values, case might not matter.

PR #134, opened 2016-04-19
Remove "char" attribute

aantron:
This attribute is deprecated in HTML5, and was reportedly not well-supported by browsers before that.

It could be marked deprecated instead, but it seems the argument for removing it entirely is stronger than even for `cellspacing`/`cellpadding`.


Drup:
The other one had bogus typing that made them not even usable. It's not the case for this one, so I'll vote for deprecation instead.

aantron:
Updated.

PR #135, opened 2016-04-19
Element and attribute reflection fixes

aantron:
a_mime_type is usable in `<object>`, `<embed>`, `<area>`, and `<link>` elements, where it corresponds to the HTML attribute "type".


aantron:
Actually, don't merge this yet. I want to add several similar changes.

aantron:
Ok, that's everything I am currently aware of, w.r.t. to reflection in the `Html` signatures.

Drup:
Thanks! I guess it's bad to have multiple overlapping reflected elements ?

aantron:
What do you mean by overlapping?

PR #136, opened 2016-04-20
Rename menu ?child to ?children

aantron:
Admittedly a nit, but it bugged me that the child list was named `child`.


Drup:
We renamed various things, so we might as well change this one too. cc @vasilisp, pretty sure that's going to break eliom.

vasilisp:
OK, thanks for the heads-up. Eliom (and JSOO) master still uses TyXML 3.6. I will rebase/update/PR the patch for TyXML master ASAP.

PR #137, opened 2016-04-20
Svg fixes

aantron:
See commit messages and diffs.

### DIFF
--- a/ppx/ppx_attribute_value.ml
+++ b/ppx/ppx_attribute_value.ml
@@ -376,36 +376,33 @@ let transform =
       | "matrix" ->
         begin match spaces_or_commas_ float loc "matrix" values with
         | [a; b; c; d; e; f] ->
-          [%expr Svg_types.Matrix
-            ([%e a], [%e b], [%e c], [%e d], [%e e], [%e f])]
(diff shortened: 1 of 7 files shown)


ppx/ppx_attribute_value.ml
```diff
@@ -335,30 +335,20 @@ let offset =
   let bad_form name loc =
     Ppx_common.error loc "Value of %s must be a number or percentage" name in
 
-  let regexp = Re_str.regexp "\\([-+0-9eE.]+\\)$\\|\\([0-9]+\\)%" in
+  let regexp = Re_str.regexp "\\([-+0-9eE.]+\\)\\(%\\)?" in
```
Drup: This should really be written with the `{| |}` form.

Drup:
Nice, thanks !  We need to refactor a bit the testsuite, test_ppx is absorbing everything.

Drup:
@vasilisp There is a change to svg's wrapped functions hidden in there.

ppx/ppx_attribute_value.ml
aantron: I have a branch from weeks ago where I switched to Re combinators anyway. Can rebase, update, and PR, if you want. It will be much easier to be sure that it's correct, now that there is a test suite.

PR #138, opened 2016-04-22
Fix location handling in the PPX

aantron:
This should fix locations. See the comment in the diff for an overview of the approach. Basically, this does a "dry run" pass through the input stream, in which Markup.ml decodes and assigns locations to characters (and normalizes newlines), but does not parse further. The pass logs these locations at the beginning of each string in the quotation, and uses that to map Markup.ml locations to OCaml.

This will only work with the `input-stream` branch of Markup.ml. If we keep this approach, then I will merge that into `master`.

I chose to hardcode the encoding to UTF-8 again, because the dry run doesn't try to infer the encoding.

### DIFF
--- a/ppx/ppx_tyxml.ml
+++ b/ppx/ppx_tyxml.ml
@@ -31,8 +31,8 @@ module Loc = struct
       loc_end = shift loc_end xend ;
     }
 
-  (** Returns the real (OCaml) location of a string, taking delimiters into
-      account. *)
+  (** Returns the real (OCaml) location of the content of a string, taking
+      delimiters into account. *)
   let string_start delimiter loc =
     let delimiter_length = match delimiter with
       | None -> 1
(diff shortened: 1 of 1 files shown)


ppx/ppx_tyxml.ml
```diff
@@ -192,30 +266,8 @@ let ast_to_stream expr =
...
-      end
-      else begin
-        offset := 0;
-        items := rest;
-        next ()
-      end
-  in
-
-  Markup.fn next, (fun x -> !current_adjust_location x)
+  Markup.string (String.concat "" (List.map fst strings)),
```
Drup: I would prefer to keep the version with `next`. `String.concat` here doesn't sound like a fabulous idea.

aantron: Can you be more specific about the meaning of *fabulous*? Are you worried about the total length of the input, copying it, or what?

Drup: > the total length of the input, copying it

Both. :)

ppx/ppx_tyxml.ml
```diff
@@ -49,29 +49,105 @@ module Loc = struct
...
+        | ((noted_markup_location, _) as loc)::rest ->
+          if Markup.compare_locations
+               noted_markup_location given_markup_location > 0 then best
+          else bounded_maximum (Some loc) rest
+      in
+
+      let preceding_markup_location, preceding_ocaml_position =
+        match bounded_maximum None location_map with
+        | Some loc -> loc
+        | None -> failwith "impossible"
```
Drup: Either use `assert false` (which will give us the assert location) or use a better error message.

ppx/ppx_tyxml.ml
aantron: Specifically what is a problem due to `String.concat`, that isn't already a bigger problem due to the rest of the surrounding code? The string length limit on 32-bit platforms?

Drup: There is that, and the fact that it's completely unnecessary. We don't need to copy this string, `next` is easy to define and doesn't make the surrounding code more complex.

aantron: Sounds like an argument for `String.concat`. `String.concat` doesn't require a definition, using it is way simpler than defining `next`, and is much simpler for a reader to understand. It is `next` that is completely unnecessary, except in case someone pastes several megabytes of markup into a PPX quotation. That doesn't seem like a realistic scenario. All other costs of `String.concat` are dwarfed by HTML parsing.

If you insist, I'll restore it, but I think it's completely pointless. The only reason for `next` in the past was that it caused side effects, which is why there is now a `next_byte` in `make_location_map`.

ppx/ppx_tyxml.ml
```diff
@@ -49,29 +49,105 @@ module Loc = struct
...
+
+    (* The function proper which translates Markup.ml locations into OCaml
+       locations. *)
+    fun given_markup_location ->
+      let rec bounded_maximum best = function
+        | [] -> best
+        | ((noted_markup_location, _) as loc)::rest ->
+          if Markup.compare_locations
+               noted_markup_location given_markup_location > 0 then best
+          else bounded_maximum (Some loc) rest
```
Drup: If I understand correctly, `best` here is the last markup location that was smaller ? Could you change the name to reflect that ?

ppx/ppx_tyxml.ml
Drup: > except in case someone pastes several megabytes of markup into a PPX quotation

Pretty sure that's going to happen at some point when I add the converter `.html/.mustache -> ocaml` ....

ppx/ppx_tyxml.ml
Drup: I think you can rewrite it in a simpler way by doing a match on two elements of the list, such that the given markup location is in between.

Drup:
I like the new approach much better. The description is a bit difficult to understand though. You could simply say that you first build a table associating the location of each beginning of string to its markup location, using markup's own stream facility. The translation from markup to ocaml locations then simply use this table to shift things around appropriately.

Drup:
Hum, could you add a pin to the appropriate version of markup in the jenkins file here: https://github.com/ocsigen/tyxml/blob/master/.jenkins.sh#L2 ? That should make jenkins happier.

PR #139, opened 2016-04-22
Support nested let%html... in...

aantron:


Drup:
Oh ! Missed that, thanks.

PR #143, opened 2016-04-23
More SVG fixes

aantron:


Drup:
Wow, thanks! :tada:

Drup:
One of the test seems to fail.

aantron:
Not with current Markup.ml. Already pinning it in Jenkins now.

aantron:
I'll just make a new PR for the Jenkins pin, then I will rebase both PRs over it after merging that. BTW, TyXML has required a newer Markup.ml than in OPAM this whole time, we just weren't testing enough to see it... :)

PR #148, opened 2016-05-04
README fixes

aantron:
I couldn't resist adding "or [PPX version of `<a>` sample]," as a suggestion. However, if you really don't want it, I will remove it.


Drup:
Thanks!

PR #150, opened 2016-05-07
Add more pointers to the ppx stuff, and ease install.

Drup:
Follow up of several discussions with @antron


Drup:
See the result [here](https://github.com/ocsigen/tyxml#tyxml) and [here](http://ocsigen.org/tyxml/dev/manual/intro).

PR #154, opened 2016-05-07
Require Markup.ml 0.7.2.

aantron:
I also removed the Jenkins pin on the Markup.ml development branch, so this should be merged after https://github.com/ocaml/opam-repository/pull/6472 and the new `markup` is installable from OPAM. I'll bump this thread when that's so.


aantron:
This is ready.

PR #152, opened 2016-05-07
Add a contributing file.

Drup:

PR #155, opened 2016-05-07
Documentation nit list.

aantron:
As requested.

Proofread the 3 files in the diff. Based on #152.

I think for really proper usage, you should capitalize PPX where it is not the name of an option (i.e. `-ppx`). The lowercase usage seems to have evolved from people dropping the `-` from `-ppx`. We have both in TyXML right now, e.g. change log has PPX, README has ppx.

Also, I think you should delete the "Ocaml libary to..." part of the tag line in GitHub :)

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -2,7 +2,7 @@
 
 Thanks for contributing!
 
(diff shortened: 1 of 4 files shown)


Drup:
Great! Thanks a lot.

PR #158, opened 2016-05-27
Flush error message formatter in the PPX.

### LINKED ISSUE #157: tyxml.ppx error messages come up as empty
Reproducing with tyxml 4.0.0 and ocaml 4.03.0+fp+flambda (both from opam):

    [14:54] aiee(pts/48):/tmp% echo 'open Tyxml let%html a = "<html>error</html>"' > foo.ml && ocamlfind ocamlopt -o foo -package tyxml.ppx foo.ml -linkpkg
    File "foo.ml", line 1, characters 31-32:
    
    [14:54] aiee(pts/48):/tmp% 

Expected behavior: error message contents is displayed.

aantron:
Resolves #157. cc @eras

With this change, the output is (I reformatted the example slightly):

```
File "foo.ml", line 2, characters 20-21:
Error: head element must have exactly one title child element
```

The error itself may seem counterintuitive, but the spec says to silently insert `<head>` and `<body>` elements, then put `error` into the body. The PPX then requires `<head>` to have a title, but it doesn't. Perhaps we should address implicit elements in error message better in the future.

### DIFF
--- a/ppx/ppx_common.ml
+++ b/ppx/ppx_common.ml
@@ -65,7 +65,9 @@ let error loc ppf =
   let buf = Buffer.create 17 in
   let fmt = Format.formatter_of_buffer buf in
   Format.kfprintf
-    (fun _ -> Location.raise_errorf ~loc "%s" (Buffer.contents buf))
+    (fun _ ->
+      Format.pp_print_flush fmt ();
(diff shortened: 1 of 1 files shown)


eras:
I don't mind if the error is a bit obtuse (though it seems fine to me), only that there was none :).

Thanks!

PR #166, opened 2016-11-21
Support for Uutf 1.0.0.

dbuenzli:
Do not merge before Uutf 1.0.0 is released (will happen soon) otherwise you'll need a a pin on `uutf`. Adds a dependency on the `uchar` compatibility package to the project.

### DIFF
--- a/lib/xml_print.ml
+++ b/lib/xml_print.ml
@@ -154,36 +154,39 @@ module Utf8 = struct
       Uutf.String.fold_utf_8
         (fun _ _ d ->
            match d with
-           | `Uchar 34 ->
-               Buffer.add_string buffer "&quot;"
-           | `Uchar 38 ->
-               Buffer.add_string buffer "&amp;"
-           | `Uchar 60 ->
-               Buffer.add_string buffer "&lt;"
-           | `Uchar 62 ->
-               Buffer.add_string buffer "&gt;"
-           | `Uchar code ->
-               let u =
-                 (* Illegal characters in html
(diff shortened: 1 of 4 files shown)


lib/xml_print.ml
```diff
@@ -154,36 +154,39 @@ module Utf8 = struct
...
+                       (code <= 31 && code <> 9 && code <> 10 && code <> 13)
+                       (* B. DEL + control C1
+                          - invalid in html
+                          - discouraged in xml;
+                          except 0x85 see http://www.w3.org/TR/newline
+                          but let's discard it anyway *)
+                       || (code >= 127 && code <= 159)
+                       (* C. UTF-16 surrogate halves : already discarded
+                          by uutf || (code >= 0xD800 && code <= 0xDFFF) *)
+                       (* D. BOM related *)
```
dbuenzli: It looks like a failed attempt at excluding [non characters](http://www.unicode.org/faq/private_use.html#nonchar4). But for example if you need to match what is [allowed](https://www.w3.org/TR/xml/#charsets) by XML this definition is too large.

dbuenzli:
Uutf 1.0.0 was released https://github.com/ocaml/opam-repository/pull/7896

lib/xml_print.ml
Drup: @vouillon, @hhugo, do you remember ? I don't have time to look at it for a few days.

Drup:
@dbuenzli Thanks a lot!

jrochel:
Bump! (with all due courtesy)

PR #171, opened 2017-05-07
Added a_clip_path attribute.

sagotch:
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/clip-path


Drup:
This doesn't seem to build. Could you add a link to the mozilla and/or the standard in the documentation (I know it's mostly not done, but I would like people to do it for new elements).

Additionally, you will need to add things for the ppx to work properly (at the very least, a test). If you can't figure it out, I can look at it later.

sagotch:
Hum... I'll look at it.

Drup:
The clip path attribute is missing far too many cases. I'll let you revisit it when you have time. When that is done, I can help you with the ppx part. In the meantime, I'm closing.

PR #172, opened 2017-05-14
Add support for the `role` attribute.

### LINKED ISSUE #165: Missing `a_role`
`a_role` seemed to be missing for the `a` element.

Armael:
See https://www.w3.org/TR/role-attribute/ and also https://www.w3.org/TR/wai-aria/

Fixes #165

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,6 @@
+* Add support for the `role` attribute (contribution by Armaël Guéneau)
+  (see https://www.w3.org/TR/role-attribute/)
+
 ===== 4.1.0 ====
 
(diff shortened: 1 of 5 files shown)


Armael:
I'm not 100% sure if `role` can actually be put on any element, given my inability to decipher the w3c doc page linked in the previous message.

Drup:
The documentation seems ill-formated. From the log, it seems ocamldoc doesn't like your @see tag.

Armael:
Indeed. I fixed the documentation and added a few more links.

Drup:
The w3c documentation is never a masterpiece of clarity, but in this case, I also think it should be available on all elements. Thanks!

PR #174, opened 2017-05-15
Add basic support for aria-* attributes

Armael:
See https://www.w3.org/TR/wai-aria-1.1/#states_and_properties

The specification indicates that many aria- attributes are only valid for
elements of a given "role" (as specified by the "role" attribute). Here, we do
not try to encode these invariants in tyxml types (it is considered not worth
the trouble). A single [a_aria] function is thus provided, with [a_aria "foo"]
corresponding to the "aria-foo" attribute.


Drup:
Nice. Can you add some tests (for ppx, notably).

Armael:
I added a ppx test, I don't know what other kind of tests I could do...

Drup:
That was enough testing for me, Thanks!

PR #183, opened 2017-07-25
Html sigs avoid variants

### LINKED ISSUE #182: Html_sigs.T should avoid using Variants
The [Html_sigs.T](https://ocsigen.org/tyxml/dev/api/Html_sigs.T) module is inconsistent in its use of the types defined in [Html_types](https://ocsigen.org/tyxml/dev/api/Html_types). Most of the time it makes use of these, but in some occasions it redefines the types.

In my opinion this is bad for several reasons:
- Code duplication (the types are always already defined in Html_types);
- Error-prone (Indeed, I spotted two inconsistencies between Html_sigs.T and Html_types);
- Less readable;
- Bloats the Html_sigs.T interface.

I compiled a list of the places where variants are used instead of dedicated types.

Elements signatures not using their associated attrib type:

iframe
object
embed
area
fieldset
optgroup
details
command
menu
link

Elements signatures not using their associated content type:

footer (uses flow5_without_header_footer)
header (uses flow5_without_header_footer)
bdo (uses phrasing)
iframe
fieldset (uses flow5)
optgroup
details (uses flow5)

Elements signatures not using their associated output type:

footer
header
bdo
a
del
ins
iframe
object (!! the types do not agree on Object_interactive variant)
embed
area
map (!! the types do not agree : Map versus A)
fieldset
datalist
optgroup
details
command
menu
link

shepard8:
This is the easy part of #182. All the types defined in Html_sigs.T that were exactly as defined in Html_types have been replaced by the corresponding type (defined in Html_types).

There remain two problems, that should be addressed in separate issues:

- The area element allows some Download attribute in Html_types, but not in Html_sigs.T;
- The object element allows some Object_interactive output in Html_types, but not in Html_sigs.T.

Hence, this closes #182.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,7 @@
+* Various fixes in the Html_sigs.T module (contribution by Fabian Pijcke):
+  * Fixed the map element function signature.
+  * The elements functions now (almost) all make use of the types defined in Html_types, rather than redefining them.
+  * Html_sigs.T.fieldset now takes [< legend] elt wrap as optional argument rather than legend elt wrap.
 * Add basic support for `aria-*` attributes (contribution by Armaël Guéneau)
   (see https://www.w3.org/TR/wai-aria-1.1/#states_and_properties)
 * Add support for the `role` attribute (contribution by Armaël Guéneau)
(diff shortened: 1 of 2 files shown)


Drup:
Nice work! On principle I agree with all the changes. I will not have time to review carefully in the next few days, but I'll do it after that.

Drup:
*/me wakes up after a long sleep*

PR #187, opened 2017-08-20
Simplistic indentation for the Format-based printer.

Drup:
The goal is not to have a great indentation engine, but something that is sufficient to be readable. There are many much more sophisticated pretty-printers for html out there. :)

Here is the output for the small website example:

```html
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml">
 <head><title>Your Cool Web Page</title>
  <link rel="stylesheet" href="home.css"/>
 </head>
 <body><div id="image_box"></div>
  <div id="content_box">
   <nav>
    <ul class="links_bar" id="links_bar"><li id="home_click">My Musings</li>
     <li id="about_click">About Me</li><li id="blog_posts_click">Blog</li>
     <li id="hackathons_click">Hackathons</li>
    </ul>
   </nav><div id="payload"><div><h2>Hello Coder</h2></div></div>
   <footer id="footer_box">
    <p>This site was made with <a href="http://ocaml.org">OCaml</a> and 
     <a href="https://www.gnu.org/software/emacs/">emacs</a>
    </p>
   </footer>
  </div><script src="main.js"></script>
 </body>
```

I haven't benchmarked yet.

### DIFF
--- a/examples/basic_website/site_html.ml
+++ b/examples/basic_website/site_html.ml
@@ -58,7 +58,7 @@ let emit_page (name, page) =
   Printf.printf "Generating: %s\n" name ;
   let file_handle = open_out name in
   let fmt = Format.formatter_of_out_channel file_handle in
-  pp () fmt page;
+  Format.fprintf fmt "%a@." (pp ()) page;
   close_out file_handle
 
 let () = List.iter emit_page pages
--- a/examples/basic_website/site_html.ml
+++ b/examples/basic_website/site_html.ml
@@ -58,7 +58,7 @@ let emit_page (name, page) =
   Printf.printf "Generating: %s\n" name ;
   let file_handle = open_out name in
   let fmt = Format.formatter_of_out_channel file_handle in
-  Format.fprintf fmt "%a@." (pp ()) page;
+  Format.fprintf fmt "%a@." (pp ~indent:true ()) page;
   close_out file_handle
 
 let () = List.iter emit_page pages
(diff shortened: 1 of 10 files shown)


Drup:
As I was afraid, doing it with boxes and format wreck performances. :/

Using the `emitbig` test:
- Time before:  0.046070
- Time after: 0.217038

aantron:
Not sure about the right way to do this, but it would be very useful for unit testing HTML output, when tests are based on line-by-line diffs. Performance is not a factor for this use case, as long as it is not absolutely awful.

vouillon:
Note that Format will output ellipses when the box nesting is too deep.

Drup:
@vouillon Ugh I forgot about that feature.
That being said, the default maximum number of *nested* boxes is max_int, if you have HTML that deep, you will break plenty of things before reaching printing.

vouillon:
It used to be much lower, but indeed, it has been raised some time ago...
https://github.com/ocaml/ocaml/commit/95f084e63db67dc477abebd5396f02e0c15138b3

Drup:
I added an optional argument to the `pp` function: `?indent:bool` that inform if the printer should output boxes and break hints. The final output when indentation is enable is the same as shown above, while the output is as before this patch when indentation is disabled.

Perfs before:
```
% ./emitbig.native
Time:  0.056289
```

Perfs after:
```
% ./emitbig.native
Time:  0.061137
% ./emitbig.native indent
Time:  0.275601
```

Output with indentation is still slow, but output without indentation is almost as fast as before. So I'm pretty happy. The example websites have indentation enabled.

Test fails at the moment due to https://github.com/aantron/markup.ml/issues/32 . I'll wait for that before merging.

PR #188, opened 2017-11-17
Use omp and driverify

Drup:
@aantron, @let-def : I wouldn't mind a review there. :)

### DIFF
--- a/ppx/.merlin
+++ b/ppx/.merlin
@@ -1,0 +1,2 @@
+REC
+FLG -open Migrate_parsetree -open Ast_403 -open Ppx_tools_403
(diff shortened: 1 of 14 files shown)


aantron:
Looks probably reasonable, but I have no experience with using OMP and Ocamlbuild/OASIS.

ppx/.merlin
```diff
@@ -0,0 +1,2 @@
+REC
+FLG -open Migrate_parsetree -open Ast_403 -open Ppx_tools_403
```
mseri: Do you mean 405 here?

Drup:
@let-def You mean having a 406 version ? In the long term, sure. Right now, as long as it can run on 406 (which it should, thanks to omp), it's not a problem.

ppx/.merlin
Drup: Yes, thanks!

PR #189, opened 2017-11-20
Quick-and-dirty patch to compile with OCaml 4.06

jaapb:
The ppx syntax extension doesn't compile with OCaml 4.06 (the expected type of reflected_variants in ppx/ppx_reflect.ml is string * (string * string list) rather than string * (string Location.loc * string Location.loc list)). Here is a quick patch to fix that - it can probably be implemented more efficiently when generating the content, but I didn't have the time to delve into the code that deeply.

### DIFF
--- a/ppx/ppx_reflect.ml
+++ b/ppx/ppx_reflect.ml
@@ -448,7 +448,8 @@ let emit_module () =
     ] else []
   end @
 
-  List.map Combi.(let_ AC.pvar (tuple2 str (list str))) !reflected_variants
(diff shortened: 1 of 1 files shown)


Drup:
Thanks! 

That break compatibility with other versions, doesn't it ? So we can't merge as it is. 

We should be able to do something not-too-complicated by writing a slightly different function depending on the OCaml version. This should be easy with OMP, in particular. I will tackle that just after #188

jaapb:
OK, cool - you're quite right the patch shouldn't be merged as-is, thanks for looking into it.

Drup:
Fixed by #188

PR #193, opened 2017-12-29
Allow restricted flow content in dt elements

aantron:
From https://www.w3.org/TR/html5/grouping-content.html#the-dt-element:

> Content model:
Flow content, but with no header, footer, sectioning content, or heading content descendants.

By contrast, TyXML has:

https://github.com/ocsigen/tyxml/blob/820d3c6b3d795aec948314e026ec0c9484b54762/lib/html_types.mli#L1397

This isn't quite restrictive enough, because, e.g., a `div` could have a nested `h2`, but I am under the impression that is already an issue in TyXML.

### DIFF
--- a/lib/html_types.mli
+++ b/lib/html_types.mli
@@ -1105,6 +1105,12 @@ type flow5_without_sectioning_heading_header_footer_address =
        flow5_without_media) transparent
   ]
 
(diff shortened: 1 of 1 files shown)


Drup:
I'm merging, as I'm quite convinced it's correct, but I wouldn't mind some tests if/when you have time.

PR #197, opened 2018-04-09
Dunification

Drup:
I think we can do it this time! :)
It's not completely finished: I need to look at the ppx failing tests, the ocsigen doc generation and sort out CI. But otherwise, it should be on the right track.

cc @rgrinberg (and @aantron ?), I would appreciate review!

### DIFF
--- a/test/jbuild
+++ b/test/jbuild
@@ -11,6 +11,7 @@
 
 (alias
  ((name runtest)
+  (deps (main_test.exe))
   (action (run ${exe:main_test.exe}))
 ))
 
@@ -25,11 +26,25 @@
 (rule
  ((targets (html_fail.result))
   (deps   (ppx.exe html_fail.ml))
-  (action (system "TERM=dumb ${OCAML}  -I ../implem/.tyxml.objs/ -noinit -noprompt -ppx './${exe:ppx.exe} --as-ppx' < html_fail.ml 2>&1 | tail -n +3 > ${@}"))
+  (action (system "TERM= ${OCAML} -I ../implem/.tyxml.objs/ -noinit -noprompt -ppx './${exe:ppx.exe} --as-ppx' < html_fail.ml 2>&1 | tail -n +3 > ${@}"))
 ))
 
 (alias
  ((name   runtest)
   (deps   (html_fail.result html_fail.expected))
-  (action (diff html_fail.result html_fail.expected))
+  (action (diff html_fail.expected html_fail.result))
 ))
+
+;; Emitbig
+
+(executable
+ ((name        emitbig)
+  (libraries  (tyxml unix))
+  (modules (emitbig))
+))
+(alias
+ ((name   runtest)
+  (deps (emitbig.exe))
+  (action (run ${exe:emitbig.exe}))
+))
+


tools/jbuild
```diff
@@ -0,0 +1,13 @@
+(jbuild_version 1)
+
+(library
+ ((name tyxml_tools)
+  (public_name tyxml.tools)
+  (wrapped false)
+  (modules tyxml_name)
+  (libraries (bytes))))
```
rgrinberg: There's no need to ever use the `bytes` lib with jbuilder. This library is only useful for building on `<4.02` which you can't do anyway with jbuilder. On the other hand, it forces you to declare a dependency on `base-bytes` which depend on `findlib`. A big negative IMO.

test/jbuild
```diff
@@ -0,0 +1,35 @@
...
+(rule
+ ((targets (html_fail.result))
+  (deps   (html_fail.ml))
+  (action (system "TERM=dumb ${OCAML} -noinit -noprompt -ppx ./${exe:ppx.exe} < ${<} 2>&1 | tail -n +2 > ${@}"))
+))
+
+; (alias
+;  ((name   runtest)
+;   (deps   (html_fail.result html_fail.expected))
+;   (action (run diff -dEbZBt html_fail.result html_fail.expected))
```
rgrinberg: If you revive this test, take a look at the `diff` action. It streamlines the definition of such tests.

META.tyxml.template
```diff
@@ -0,0 +1,2 @@
+# JBUILDER_GEN
+archive(byte,toploop) += "tyxml_top.cma"
```
rgrinberg: Wouldn't it be better to use `requires` and `tyxml.top` as opposed to `archive` here? It's a bit more general.

Drup: The goal of this is to auto-load `tyxml.top` in the toplevel when `tyxml` is #required. Do you have another way of doing that?

test/jbuild
Drup: I know about it, but it doesn't support custom flags. :/

tools/jbuild
Drup: Fair enough!

META.tyxml.template
rgrinberg: `requires(byte, toploop) += "tyxml.top"` should do the same thing but also handle dependencies if your toplevel has dependencies. Anyway, I'm not too sure what's best here. None of it is gonna work with `$ jbuilder utop` anyway :/

test/jbuild
rgrinberg: There's `--diff-command` to set it for your project. But you're right, this should be customizable per action as well. I'd open a ticket regarding this.

META.tyxml.template
Drup: Ah, I see. Yes, your version is better. I should probably open a ticket about better support for this.

Drup:
@rgrinberg I fixed the toplevel test in a way that is utterly disgusting, but I don't know how to do better. Do you have any propositions ?

rgrinberg:
What you can try and do is manually create a driver executable using your local ppx and then see if you can massage the flags to make it output what you want.

I managed to do this in my ppx_getenv2 example project: https://github.com/rgrinberg/ppx_getenv2/blob/master/test/jbuild#L16

But I used ppx_driver. You can probably get it to work with omp.driver as well.

Drup:
@rgrinberg the ppx is not the problem. The issue is that the code that is executed by the toplevel should have visibility over the `tyxml` library. So, it needs a `-I <the path where .cmis are`. I couldn't find a way to do that propery, so the current version just hardcode the path inside `_build`.

rgrinberg:
Oh, I see that's a bit shitty. However, can't you build a custom toplevel as an executable, include `tyxml` normally as a library and then use that to execute your code? You can the preprocessing of the code using the manual driver as I showed.

Drup:
Building a custom toplevel that `-linkall` against a library only handles the runtime part (cmo and cmas). It doesn't bundle the cmis for typechecking. For that, you still need `-I`.

rgrinberg:
Oh yeah, I remember having to do that to make `$ jbuilder utop` work. Sounds like what you need are some decent rules for making toplevels. We should generalize `$ jbuilder utop`, I suppose. First make the utop dependency optional, and also allow the user to refer to it as a target.

PR #200, opened 2018-04-18
Make the ppx work with wrapped APIs.

Drup:
I was somehow convinced that this was working already, but apparenty it didn't.
I extended the test cases to cover wrapped APIs and I fixed a few bugs.

After this patch, this works:

```ocaml
open Tyxml_js.R

let%html make title content =
  {|<html>
     <head>
       <title>|}title{|</title>
     </head>
     <body>|}content{|</body>
   </html>
  |}
```

and yields a completely reactive webpage. :heart: 
```ocaml
val make :
  [< Html_types.title_content_fun ] Tyxml_js.To_dom.elt React.signal ->
  [< Html_types.object__content_fun > `PCDATA ] Tyxml_js.To_dom.elt ReactiveData.RList.t ->
  [> Html_types.html ] Tyxml_js.To_dom.elt
```

### DIFF
--- a/examples/mini_website_ppx/minihtml.ml
+++ b/examples/mini_website_ppx/minihtml.ml
@@ -15,7 +15,7 @@ let%html mypage =
      <head>
        <title>|}mytitle{|</title>
      </head>
-     <body>"mycontent"</body>
+     <body>|}[mycontent]{|</body>
    </html>
   |}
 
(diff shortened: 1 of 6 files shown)


Drup:
@vasilisp I added a test for attributes that uses wrapped functions, so I believe it should also work with eliom's shared dom (and all its fancy variations). Can you give it a shot?

vasilisp:
@Drup thanks, I promise to try it out in the next few days!

PR #202, opened 2018-04-19
Try to get travis to run the tests

Drup:


Drup:
@rgrinberg The second commit make jbuilder give this error:
```
Dependency cycle between the following files:
    _build/.aliases/default/.tyxml-files-00000000000000000000000000000000
--> _build/.aliases/default/.tyxml-ppx-files-00000000000000000000000000000000
--> _build/.aliases/default/.tyxml-files-00000000000000000000000000000000
```

But there shouldn't be any dependency cycle here. `html_fail.ml` is only used for tests. If I don't have this dependency, I will get an error if I run `jbuilder runtest -p tyxml-ppx` when tyxml is not locally compiled.

Drup:
@rgrinberg Is there a way to only run a test on some version of OCaml ? The "typechecking failure" test is ighly dependent on the output of the compiler and break regularly. I would like to only run it on one OCaml version.

rgrinberg:
>  I will get an error if I run jbuilder runtest -p tyxml-ppx when tyxml is not locally compiled.

Is this still an issue? This behavior is expected btw. `-p` is for opam installations, hence the build will look into the opam switch for all dependencies (even ones that exist locally). If you would like to run tests for a particular package you should do something like:

```
$ dune runtest <dir that contains test for that package>
```

PR #210, opened 2018-08-25
remove uchar compatability library from build

avsm:
This has been included in the OCaml standard library since
version 4.03.0, which is also the minimum version used by
tyxml.

Removing the compat eliminates a non-dune build dependency,
which in turn makes it easier to embed tyxml as part of a
larger monorepo build using a single dune invocation.

Signed-off-by: Anil Madhavapeddy <anil@recoil.org>

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -2,7 +2,7 @@ BLD=../_build/default
 SRC=..
 
 PKGS=\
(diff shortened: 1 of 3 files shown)


Drup:
Thanks! 

It seems CI is completely broken, I'll have to take a look at that ....

PR #211, opened 2018-09-03
Adding support for touch events

malthe:
This adds support for the following events:
- `ontouchstart`
- `ontouchend`
- `ontouchmove`
- `ontouchcancel`

### DIFF
--- a/lib/html_sigs.mli
+++ b/lib/html_sigs.mli
@@ -242,6 +242,12 @@ module type T = sig
   val a_onmousemove : Xml.mouse_event_handler -> [> | `OnMouseMove] attrib
   val a_onmouseout : Xml.mouse_event_handler -> [> | `OnMouseOut] attrib
 
+  (** {4 Touch events} *)
+  val a_ontouchstart : Xml.touch_event_handler -> [> | `OnTouchStart] attrib
+  val a_ontouchend : Xml.touch_event_handler -> [> | `OnTouchEnd] attrib
+  val a_ontouchmove : Xml.touch_event_handler -> [> | `OnTouchMove] attrib
+  val a_ontouchcancel : Xml.touch_event_handler -> [> | `OnTouchCancel] attrib
+
   (** {4 Keyboard events} *)
 
   val a_onkeypress : Xml.keyboard_event_handler -> [> | `OnKeyPress] attrib
--- a/lib/html_sigs.mli
+++ b/lib/html_sigs.mli
@@ -1141,6 +1141,7 @@ sig
      and type Xml.event_handler = Xml.event_handler
      and type Xml.mouse_event_handler = Xml.mouse_event_handler
      and type Xml.keyboard_event_handler = Xml.keyboard_event_handler
+     and type Xml.touch_event_handler = Xml.touch_event_handler
      and type Xml.attrib = Xml.attrib
      and type Xml.elt = Xml.elt
      and module Svg := Svg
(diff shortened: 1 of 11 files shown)


lib/svg_sigs.mli
```diff
@@ -673,7 +673,7 @@ module type T = sig
   val a_onunload : Xml.event_handler  -> [> | `OnUnload ] attrib
   val a_onzoom : Xml.event_handler  -> [> | `OnZoom ] attrib
 
-  (** {3 Javascript mouse events} *)
+  (** {6 Javascript mouse events} *)
```
Drup: Why changing the header level ?

Drup:
@vasilisp I suppose you guys already have a custom implem of that. Anything to add ? :)

vasilisp:
@Drup we clearly do use the touch events, but via `Lwt_js_events`, so I had not even noticed that they are missing in TyXML. The PR looks good (modulo your comments) :).

Drup:
I'm ... not quite sure what's happening in CI.

vasilisp:
A missing `type Xml.touch_event_handler = Xml.touch_event_handler` constraint in the `T with ...`?

Drup:
> A missing type Xml.touch_event_handler = Xml.touch_event_handler constraint in the T with ...?

Ah! Well caught.

@malthe You need to go through the various `with` constraints and every time you see events, add the touch one.

lib/svg_sigs.mli
malthe: Oh lol – I obviously didn't get what that was at all.

malthe:
@Drup I don't know if I know how to write/add PPX tests. I'm an OCaml newbie, relatively speaking ...

lib/html_sigs.mli
```diff
@@ -1141,6 +1141,7 @@ sig
      and type Xml.event_handler = Xml.event_handler
      and type Xml.mouse_event_handler = Xml.mouse_event_handler
      and type Xml.keyboard_event_handler = Xml.keyboard_event_handler
+     and type Xml.touch_event_handler = Xml.touch_event_handler
```
Drup: You should need the same thing on the svg side.

malthe: I have them there – just missed one location.

Drup:
For tests, I would like at least one test exercising events with the ppx here: https://github.com/ocsigen/tyxml/blob/master/test/test_ppx.ml#L150 . 
Otherwise, feel free to add new tests in non-ppx syntax here: https://github.com/ocsigen/tyxml/blob/master/test/test_html.ml
You can follow the patterns from the other tests.

sagotch:
> I'm ... not quite sure what's happening in CI.

It looks like a bug in `dune 1.1.0`. I see that `1.1.1` is installed, but console shows `jbuilder` invocation. Could it be using outdated code?

ppx/reflect/reflect.ml
```diff
@@ -163,6 +163,7 @@ let rec to_attribute_parser lang name = function
   | [[%type: Xml.event_handler]]
   | [[%type: Xml.mouse_event_handler]]
   | [[%type: Xml.keyboard_event_handler]] ->
+  | [[%type: Xml.touch_event_handler]] ->
```
Drup: There is a syntax error here. You still have the arrow on the line above.

Drup:
@malthe CI should now operational again. I'll merge this as soon as you have fixed the issue in reflect.ml.

Drup:
Thanks!

PR #215, opened 2018-09-24
Convert from Jbuilder to Dune

aantron:


Drup:
Does that requires a specific version of dune ?

aantron:
No, I stuck to Dune 1.0.0 features. Thanks for asking, because I forgot that the auto-generated `dune-project` gets the installed Dune version put in it (1.2 for me). I changed it to 1.0.

PR #216, opened 2018-09-24
Adapt the PPX to upcoming Markup.ml 0.9.0

aantron:
Markup.ml 0.9.0 adds a new signal, `` `Raw of string``. When the HTML and XML *writers* encounter it, they output the string payload verbatim. The *parsers* never generate this signal.

Since the PPX only uses the HTML parser, it is not affected, except that we have to add a case to ignore this signal (which the PPX will never get).

When releasing Markup, I'll update constraints/conflicts on existing TyXML releases.

### DIFF
--- a/ppx/tyxml_ppx.ml
+++ b/ppx/tyxml_ppx.ml
@@ -326,7 +326,7 @@ let markup_to_expr lang loc expr =
       let node = Common.value @@ Element.comment ~loc ~lang s in
       assemble lang (node :: children)
 
-    | Some (`Xml _ | `Doctype _ | `PI _)  ->
+    | Some (`Xml _ | `Doctype _ | `PI _ | `Raw _)  ->
       assemble lang children
   in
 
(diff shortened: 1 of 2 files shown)


Drup:
Wait, since you are using polymorphic variants and you never generate them in the parser. You don't even need to have them in the types. Why did you changed the type of both the printer and the parser ?

aantron:
Mainly to avoid a proliferation of types for various function signatures. There are already signals that aren't generated by the HTML parser, namely `` `PI `` and `` `Xml ``.

Drup:
I understand, but this seems to be this is an useless loose signature that introduces an unnecessary breaking change. Is it too late for you to reconsider ?

aantron:
No, not too late. I'll experiment with more precise types, but I suspect it will make the signature illegible for the sake of avoiding forcing the user to write lots of `:>`.

aantron:
We no longer need the `` `Raw `` signal in odoc, so I'll just put that in a branch in Markup.ml, not release it, and figure out what to do with it later. Nothing needs to be changed in TyXML.

Drup:
Ok, thanks!

PR #225, opened 2018-12-15
Whitespace in table

Julow:
Whitespaces are treated as `PCDATA` which are not accepted in `table` and `tr` elements.

~~I took example on #121 but I suspect this is not the best approach as this problem may occur in a lot of other elements. Would adding some kind of `[@@reflect.filter_whitespace]` in `html_sigs.mli` a better solution ?~~

~~While adding a test, I could not write `[table [tbody [tr [td [txt "foo"]] ; tr [td [txt "bar"]]]]]`. `table` does not accept `tbody`.~~
~~I tried doing the same for `colgroup`, `thead` and `tfoot` elements (see https://github.com/Julow/tyxml/compare/whitespace_in_tr...whitespace_in_table_full) but I had a similar problem.~~

Use the annotation added in https://github.com/ocsigen/tyxml/pull/226

### DIFF
--- a/ppx/element_content.ml
+++ b/ppx/element_content.ml
@@ -173,6 +173,8 @@ let table ~lang ~loc ~name children =
   let thead, others = partition (html "thead") others in
   let tfoot, others = partition (html "tfoot") others in
 
+  let others = filter_whitespace others in
+
   let one label = function
     | [] -> []
     | [child] -> [Common.Label.labelled label, Common.wrap_value lang loc child]
--- a/ppx/element_content.ml
+++ b/ppx/element_content.ml
@@ -173,8 +173,6 @@ let table ~lang ~loc ~name children =
   let thead, others = partition (html "thead") others in
   let tfoot, others = partition (html "tfoot") others in
 
-  let others = filter_whitespace others in
-
   let one label = function
     | [] -> []
     | [child] -> [Common.Label.labelled label, Common.wrap_value lang loc child]
--- a/test/test_ppx.ml
+++ b/test/test_ppx.ml
@@ -137,6 +137,45 @@ let basics = "ppx basics", HtmlTests.make Html.[
   [[%html "<ol>   <li>foo</li>  <li>bar</li>   </ol>"]],
   [ol [li [txt "foo"] ; li [txt "bar"]]] ;
 
+  "whitespace in tr",
+  [[%html "<tr>   <td>foo</td>    <td>bar</td>   </tr>"]],
+  [tr [td [txt "foo"] ; td [txt "bar"]]] ;
+
+  "whitespace in table",
+  [[%html "<table>    <tr><td>foo</td></tr>   <tr><td>bar</td></tr>   </table>"]],
+  [table [tr [td [txt "foo"]] ; tr [td [txt "bar"]]]] ;
+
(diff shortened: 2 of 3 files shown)


ppx/element_content.ml
```diff
@@ -181,6 +181,8 @@ let table ~lang ~loc ~name children =
   let thead, others = partition (html "thead") others in
   let tfoot, others = partition (html "tfoot") others in
 
+  let others = filter_whitespace others in
```
Drup: That's going to filter whitespaces everywhere else, that's not correct.

test/test_ppx.ml
```diff
@@ -137,6 +137,15 @@ let basics = "ppx basics", HtmlTests.make Html.[
   [[%html "<ol>   <li>foo</li>  <li>bar</li>   </ol>"]],
   [ol [li [txt "foo"] ; li [txt "bar"]]] ;
 
+  "whitespace in tr",
+  [[%html "<tr>   <td>foo</td>    <td>bar</td>   </tr>"]],
+  [tr [td [txt "foo"] ; td [txt "bar"]]] ;
+
+  "whitespace in table",
+  [[%html "<table>    <tr><td>foo</td></tr>   <tr><td>bar</td></tr>   </table>"]],
+  (* [table [tr [td [txt "foo"]] ; tr [td [txt "bar"]]]] ; *)
```
Drup: `table [tr [td [txt "foo"]] ; tr [td [txt "bar"]]]` should work (without a `tbody`).

Julow: Doing this I get `expecting
<table><tbody><tr><td>foo</td></tr><tr><td>bar</td></tr></tbody></table>, got
<table><tr><td>foo</td></tr><tr><td>bar</td></tr></table>.`. What did I miss ?

ppx/element_content.ml
Julow: I though it would remove whitespaces around `tr`.
`others` should contains only `tr` ?

Julow:
I'll try implementing the `filter_whitespace` annotation.

Julow:
I implemented the annotation in https://github.com/ocsigen/tyxml/pull/226 but I'm still stuck with tables in the test suite.
Also, do you prefer a new PR or a force push on this one ?

ppx/element_content.ml
Drup: Ah, you are right, I misread the code. In this case, should't whitespace be filtered directly at the beginning, before the partitions ?

test/test_ppx.ml
Drup: Hm, the combinators and the syntax should produce exactly the same thing. This sounds like a bug.

Drup:
I'll need to investigate the table issue more closely. I'm not sure what the tbody deal is.

Feel free to force push.

ppx/element_content.ml
Julow: In that case, with the filter_whitespace annotation !

Drup:
1 year and half later .. (sorry !) : `table` is split into 2 versions, `table` which only accepts elements `tr` elements directlyy, and `tablex`, which only accepts `tbody`. I fixed the tests by using `tablex` where appropriate. I'll merge as soon as CI comes back.

PR #226, opened 2018-12-16
Add `[@@reflect.filter whitespace]` annotation

Julow:
It calls `filter_whitespace` on children just before passing them to the assembler. It avoids the need to write an explicit assembler.

In the 2nd commit I replaced the assembler for `ul`, `ol` and `select` with this annotation and used it on top of `head` and `html`.

I tried to implement it as generic as possible. Adding similar annotations should be very easy.

### DIFF
--- a/lib/html_sigs.mli
+++ b/lib/html_sigs.mli
@@ -684,11 +684,13 @@ module type T = sig
   val html :
     ?a: ((html_attrib attrib) list) ->
     [< | head] elt wrap -> [< | body] elt wrap -> [> | html] elt
+  [@@reflect.filter_whitespace]
   [@@reflect.element "html"]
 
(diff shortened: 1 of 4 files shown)


Drup:
Nice job! I'll wait for CI to come back, but the implementation looks good. Now you'll just need to go through all the HTML elements (and SVG, if you are courageous) and annotate them!

Julow:
I couldn't find other elements that use `filter_whitespace`. Do you mean annotating every elements that need this behavior ? (maybe every elements that do not accept `PCDATA`)

Drup:
Yes, I mean adding the one that should have it, but do not currently.

Drup:
Ok, let's start by merging that one!

PR #227, opened 2018-12-17
Add html_of_wiki automatic doc generation to Travis CI.

leovalais:
- Moved `ocsigen.org-data/tyxml/*` into `docs`.
- CI updated to regenerate the website for each commit on `master`.


Drup:
Huh, wait, no. I don't want the auto-generated doc in the master branch.

Drup:
Both. This PR pushes a really big bunch of generated files in master. Those used to sit in another branch, and they should stay there.

leovalais:
Actually, the API files of other versions than `dev` were generated automatically, but we cannot regenerate them only with what's on master. The only way to generate all the API files inside the CI would be to checkout on the release commit of each version, run ocamldoc and collect the files. Even by doing so, I don't know if there is a guarantee to generate the exact same files.
The only ones that could be easily generated automatically for now are `dev/api/*`. I discussed this issue with @balat and this is the only "solution" we found.

Drup:
I don't mind storing them at all, I just don't want them *in the master branch*. Why can't you put them in another branch ?

leovalais:
OK. Could you just explain what cons do you see having everything in master like that?
Also, would you want `dev` in the other branch or in master?

Drup:
In your branch `how-ci`, the `/docs` directory is 76Mo. The entire rest of the files represents less than 500Ko. I don't want to impose to everyone checking out tyxml's master to download 76Mo worth of stale documentation that is only used for website generation. If it's placed in another branch, it will not get fetched by maintainers, like me.

I would prefer to have `/docs/dev` somewhere else as well, it would make things simpler: might as well have all the generated (past or present) doc at the same place.

PR #230, opened 2018-12-18
Move docs in wikidoc branch with dedicated Travis deployement to GitHub Pages.

leovalais:
Following the discussion on #227, I moved the `docs` directory in its own branch (`wikidoc`). Each commit on this branch regenerates the website.
Since the branch is dedicated to documentation, I also removed the unrelated files (sources, tests, opam, etc.).


Drup:
Much better. I wouldn't mind an explanation of the CI workflow that will be used (I would have expected the CI script to end up in master). 

One (hopefully final) request: Instead of having a branch forked of master where you remove everything, you should start an orphaned branch using `git checkout --orphan name_of_branch`,

PR #231, opened 2018-12-19
html_of_wiki and Travis CI (orphan branch)

leovalais:
Following #230, `wikidoc` is an on orphan branch.

> I wouldn't mind an explanation of the CI workflow that will be used

Travis allows multiple configuration files. The CI will run for each commit (and merge commit) on `wikidoc`. It automatically downloads and setup html_of_wiki and generates the HTML pages based on the branch's content. I've already setup the configuration for html_of_wiki for this project. 
Then, only on the `wikidoc` branch, it deploys the content to github pages. Thus the CI on `master` (and its forked branches) is unchanged.

Part of the documentation (*still a WIP*) of html_of_wiki can be found at http://ocsigen.org/html_of_wiki/2.0/manual/intro


Drup:
Good!

It's still not clear to me how you generate the API doc and push the resulting `.wiki` to `docs/dev/`. Also, I would like to know the exact things I need to do when drafting a new release (which should be documented [here](https://github.com/ocsigen/tyxml/blob/master/docs/RELEASE.md)), but you can clarify all that later.

Drup:
Quick feedback on the documentation tooling
1. Just make one single tool with multiple "verbs" (like git/opam) instead of playing clever games with acronyms. 
2. Since you seem to be working on the general tooling: keep in mind that the ultimate goal (might be in the long term, but still) is to use odoc for everything instead. So everything that you do *not* implement and replace by externally maintained things is a net plus.

Drup:
@leovalais Issue in the build it seems : https://travis-ci.org/ocsigen/tyxml/builds/470166584?utm_medium=notification&utm_source=email

leovalais:
Thanks for the feedback!
> Just make one single tool with multiple "verbs" (like git/opam) instead of playing clever games with acronyms.

It is what is planned (https://github.com/ocsigen/html_of_wiki/issues/108) but I don't think I will have enough to finish that properly before the end of my internship at BeSport. So I wrote the two scripts (the dop and quickdop things) in order to have it at least working in a CI. But you're absolutely right, it is not convenient at all.

> So everything that you do not implement and replace by externally maintained things is a net plus.

I know that html_of_wiki will eventually be replaced by odoc (for the better), but in the short term at least, there's still the need of a tool to deal with the existing documentation. I tried to keep it as simple as possible (just enough for the ocsigen projects).

leovalais:
> @leovalais Issue in the build it seems : https://travis-ci.org/ocsigen/tyxml/builds/470166584?utm_medium=notification&utm_source=email

A Github access token was missing. I restarted the build: it should work properly now.

PR #237, opened 2019-01-10
Remove documentation from master and update instructions about releases.

leovalais:
* As the docs are supposed to be on `wikidoc`, I removed them from `master` to avoid duplication.
* @drup as you asked in #231, I updated the file README.md.


Drup:
I didn't understood the doc would be *removed* from master. I though the script would copy them to wikidoc/dev, along with the (generated) wikified API doc. This way, the branch `wikidoc` would only contain generated (present and past) documentation. The hand-written one would stay in master.

I really don't like the idea of the handwritten things being moved to a branch that contains mostly generated doc. I want to keep the handwritten documentation in master.

cc @aantron , as I suppose he will have similar concerns for lwt.

Drup:
*Sight*, so the doc infrastructure is still half broken....

So there is no automatic copying of dev documentation during CI, like we used to have ? Could you at least write me a build target that assumes everything is built locally, and copy all the appropriate file to the `wikidoc` branch ? Also, I would rather you fix what's written in (docs/Release.md). As far as I understand, nothing changes for contributors.

Drup:
I just tool a look at the contributing guide

> Remember, you have to issue 2 pull requests—one for your code on master and one for its documentation on wikidoc—for your contribution to be accepted! 

Yeah, I'm sorry, from a FOSS point of view, that's broken and I don't want to deal with that.

> Make sure the documentation on wikidoc is up to date with the code on master. 

That is precisely what I want you to explain. How do I generate the dev API `.wiki` files . Is the `make wikidoc` target still appropriate ? If no, remove it. In any case, that's precisely the part I hoped you would automate ....

In general, I would have appreciate being asked what @aantron and I expect from a system like that, and explained exactly how it was going to work. Having to double-check what you do without any explanation of the pipeline, which parts are automated and which aren't is not pleasant.

aantron:
I agree that we can't, practically, ask open source contributors to make a PR to `wikidoc` to update the docs. Everything "normal" that is editable by humans should be in `master`, and the process of updating `wikidoc` and `gh-pages` should eventually be fully automated.

Drup:
Due to lack of activity and the fact that I don't see how this could progress without a significant redesign, I'm going to close this.

balat:
I don't think it needs a significant redesign. It's probably just a few
lines in a makefile, isn't it?

Le ven. 24 janv. 2020 à 16:17, Gabriel Radanne <notifications@github.com> a
écrit :

PR #239, opened 2019-05-06
Add <template> tag

slegrand45:
- [W3C](https://www.w3.org/TR/html5/semantics-scripting.html#the-template-element)
- [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template)

### DIFF
--- a/lib/html_f.ml
+++ b/lib/html_f.ml
@@ -659,6 +659,8 @@ struct
 
   let noscript = star "noscript"
 
+  let template = star "template"
+
   let article = star "article"
 
   let aside = star "aside"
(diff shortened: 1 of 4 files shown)


Drup:
Thanks, I wanted to add this one!

However, I'm not sure I agree with the typing, I'll have to study it a bit more closely (in particular, the spec and MDN seems to disagree).

Also, please add the link to mdn in the .mli :)

slegrand45:
Link to MDN added.

Drup:
In the end, the typing seems fine, thanks!

Issue #240, opened 2019-05-22
Missing attributes for iframe


thizanne:
https://developer.mozilla.org/fr/docs/Web/HTML/Element/iframe

Attributes `a_allow`, `a_referrerpolicy` and `a_srcdoc` seem to be missing.

Drup:
Indeed. I encourage you to contribute them. ;)

thizanne:
Solved in #242

PR #242, opened 2019-05-22
Add `allowfullscreen`, `allowpaymentrequest`, `referrerpolicy` iframe attributes

thizanne:
https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-iframe-element
https://html.spec.whatwg.org/multipage/iframe-embed-object.html#attr-iframe-referrerpolicy

Note that:
- Referrerpolicy is an enumerated attribute, usually translated as a variant, except is also accepts the empty string as a valid keyword. I added a `variant_or_empty` parser that can parse this kind of attribute, parameterised by the variant to which the empty string will be mapped (given as a string, for instance `"Empty"` for the ``Empty` polyvar).
- Some other elements such as `a` and `area` should accept referrer policies, but the MDN considers this as experimental. Although I did not add this support, this should not be too hard now that the attribute exists.

Thanks @Drup for the help.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,7 +9,7 @@
   and removes the ocamlfind library `tyxml.ppx` in favor of `tyxml-ppx`.
   (#197 by Drup, Rudi Grinberg and Anton Bachin)
 * Add simplistic indentation for the Format-based printer (#187 by Drup)
-* Allow the ppx to be used for more exotic tyxml instances, such 
+* Allow the ppx to be used for more exotic tyxml instances, such
   as reactive elements (#200 by Drup)
 * Add `Html.of_seq` and `Svg.of_seq`, which allow to easily import
   HTML parsed with markup in TyXML (#221 by Drup)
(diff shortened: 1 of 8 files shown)


thizanne:
See #240

Drup:
Nice work! The addition of `variant_or_empty` is interesting and I'm happy with your solution. It does mean that `referrerpolicy=empty` will be accepted in the ppx, but I don't feel terribly bad about that.

Nevertheless, I would like a ppx test for the other cases, notably to check that `things-with-dashes` are properly translated. A test with `no-referrer-when-downgrade` would do nicely.

You also need to rebase to fix the conflict in the changelog.

thizanne:
Sorry for the force-push noise, I apparently rebased against the wrong branch. I added the ppx test and rebased nicely (now).

As for the empty thing, I agree that's not ideal, but we won't have a nice typing solution to that: we would have to use an option type or similar, and I feel like this does not semantically reflect the fact that `""` is considered as a valid keyword, not the absence of a keyword. However this could possibly appear in the documentation -- unless we prefer to hide it under the carpet (or not to actually care since it's probably no big deal).

Drup:
Well, there is also the solution to write the ppx thingy by hand. But honestly, it doesn't matter. I'll merge when CI is green.

PR #243, opened 2019-05-24
Allow crossorigin attribute for script element

thizanne:
https://developer.mozilla.org/fr/docs/Web/HTML/Element/script

`a_crossorigin` happens to already be defined, with an already correct type. This PR merely adds this as an acceptable attribute for `script`.


Drup:
Looks perfect, thanks!

PR #247, opened 2019-10-02
Add tyxml-ppx-jsx

jorisgio:
Add a new ppx reusing most of the code to translate JSX extension from reason into Tyxml combinators.


Drup:
Excelent! It's something that I wanted to implement for a while.

Before I dig into the implementation properly, Can you show/add small examples of what it looks in action ?

tyxml-ppx-jsx.opam
```diff
@@ -0,0 +1,32 @@
...
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+
+depends: [
+  "ocaml" {>= "4.02"}
+  "dune" {build}
+  "alcotest" {with-test}
+  "tyxml"
+  "tyxml-ppx"
```
Drup: Does it really need the tyxml-ppx dependency too ?

ppx/jsx-ppx/tyxml_ppx_jsx.ml
```diff
@@ -0,0 +1,96 @@
...
+    | "class_" -> "class"
+    | "className" -> "class"
+    | "htmlFor" -> "for"
+    | "type_" -> "type"
+    | "to_" -> "to"
+    | "open_" -> "open"
+    | "begin_" -> "begin"
+    | "end_" -> "end"
+    | "in_" -> "in"
+    | name -> name
```
Drup: Where does that list come from ? Is it just "standard" practice in JSX ? Otherwise, we might need the HTML/CSS<->JS full translation (which is implemented somewhere in jsoo).

ppx/jsx-ppx/register/dune
```diff
@@ -0,0 +1,6 @@
+(library
+ (name tyxml_ppx_jsx_register)
+ (public_name tyxml-ppx-jsx)
```
Drup: I would prefer "tyxml_jsx" instead.

jorisgio:
Sorry for keeping quiet :) i'm waiting for this to get more adopted at work to see some usage in the wild before moving the PR out of draft

tyxml-ppx-jsx.opam
jorisgio: In ideal, no. But opam/dune will not install the common lib for jsx ppx because it is a subpackage of tyxml-ppx. Maybe it is better to move the ppx shared lib into tyxml itself ?

ppx/jsx-ppx/tyxml_ppx_jsx.ml
jorisgio: https://github.com/cxa/ppx_bsx/blob/master/src/ppx_bsx_lib.ml#L16 i basically took a mix from this one and this one https://github.com/bikallem/jsx-ssr/blob/8bf3101d1b6ff71c96c3a1f8462cd9e137dfc8b2/lib/Jsx.re#L62. Unfortunately there does not seem to be any official convention, just loose common way of renaming :/

Khady:
Here is an example of how we try to use this new jsx syntax. I'm porting an existing template we have, from glade (a php thing). Slightly edited to post here. And a lot of small details have to be commented because it was not html5 compliant. It still doesn't compile, but it gives an idea. This is the smallest thing we have. Most of our other templates are much more complex.

```reason
open Tyxml;

let template = (~user_name) => {
<html>
<head>
    <title>"Welcome to our super product!"</title>
    /* <meta http-equiv="Content-Type" content="text/html; charset=utf-8"> */
</head>
<body /* bgcolor="#fff" */ /* leftmargin="0" */ /* topmargin="0" marginwidth="0" marginheight="0" */>
    <table /* border="0" */ /* width="100%" */ /* height="100%" */ /* cellpadding="0" cellspacing="0" */>
        <tr>
            <td /* align="left" */ /* valign="top" */ style="font-family: arial,sans-serif;">
                "Hi "user_name","/*<br>*//*<br>*/

                {|Thanks for purchasing our product.|}/*<br>*//*<br>*/

                <strong>"Access your course here:"</strong>/*<br>*/
                <a href="https://could-it-be-the-ocaml-mooc.com">"click now"</a>/*<br>*//*<br>*/


                "Enjoy, and don’t hesitate to get in touch if you have any feedback or questions for us ;)"
                /*<br>*//*<br>*//*<br>*/


                "Cheers,"/*<br>*//*<br>*/

                "The author"/*<br>*/
                <a href="https://personal-page.com">"profile"</a>/*<br>*/
                "Head of Marketing"/*<br>*//*<br>*/

                "some signature"
            </td>
        </tr>
    </table>
</body>
</html>
}

```

Drup:
Why doesn't this compile ? Also, lot's of it look HTML compliant, so it should work.

Khady:
```
File "emailer/blog.re", line 10, characters 4-940:
Error: This expression has type
         ([> Html_types.tr ] as 'a) Tyxml_html.elt list
       but an expression was expected of type
         ([< Html_types.tablex_content_fun ] as 'b) Tyxml_html.elt list
       Type 'a Tyxml_html.elt is not compatible with type 'b Tyxml_html.elt 
       Type 'a = [> `Tr ] is not compatible with type 'b = [< `Tbody ] 
       The second variant type does not allow tag(s) `Tr
```

It is probably a legitimate error. I just don't have time to fiddle with it right now.

Khady:
The error was simply that `<tr>` cannot go right inside the `<table>` and requires the `<tbody>` in between.

Khady:
What would you like us to add to make this PR ready for a proper review and merge?

I tried to see how I could add some tests. But except including some examples and checking that they compile, I don't know what is possible. Is it a problem if this package depends on reason for its tests?

Drup:
> I tried to see how I could add some tests. But except including some examples and checking that they compile, I don't know what is possible. Is it a problem if this package depends on reason for its tests?

I would like both small and big examples before considering merge. You can look at the existing tests. I don't mind having some optional tests depending on reason (even if ths jsx package itself doesn't).

I would also like to make sure it works with non-textual tyxml APIs (for example, Tyxml_js from jsoo). How was the reception from other devs at ahrefs ?

ulrikstrid:
I'm doing a small project and would love to use this to generate static HTML.

What can I help do to move this forward?

Drup:
I was partially waiting for feedback from the ahrefs people. It seems this works well, and doesn't need to be improved, so we can proceed! To give concrete steps:

1. Translate the existing examples to this new syntax
2. Add tests, probably by copying/extending the existing ppx tests
3. Make sure this works with `Tyxml_js`. I will do that myself while reviewing the code.

I was going to make a new release soon, so if someone moves 1. and 2. forward, it'll probably go in. :)

Khady:
1. is happening in https://github.com/ahrefs/tyxml/pull/1 thanks to @ulrikstrid

ulrikstrid:
2. is hopefully done here: https://github.com/ahrefs/tyxml/pull/2

Drup:
Let's pursue the discussion on #254

PR #249, opened 2020-01-05
Add data-* attribute to SVG

slegrand45:
Hello,

This is an attempt to add data-* attribute to SVG. At this time, that doesn't work with ppx. Any clue to fix it is welcome!

[https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/data-*](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/data-*)

### DIFF
--- a/lib/svg_f.ml
+++ b/lib/svg_f.ml
@@ -244,6 +244,8 @@ struct
 
   let a_id = string_attrib "id"
 
+  let a_user_data name = string_attrib ("data-" ^ name)
+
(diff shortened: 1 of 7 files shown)


Drup:
Thanks! This is a very good idea.

I'm pretty sure this is handled properly on the HTML side, so you can take a look at how it's done there. :)

Drup:
I fixed and merged in #253, thanks!

PR #251, opened 2020-01-10
Update explanatory license text

jamesjer:
The actual license text in LICENSE is the current LGPL text, and expands the first 'L' as Lesser.  The text explaining the license exception at the top, on the other hand, currently uses the obsolete expansion of 'L' as Library.  The first commit changes the license exception text to match the license.

The second commit updates the zip code for the FSF in the source files to match the one in LICENSE.  It is currently the zip code for the old Temple Place address.


Drup:
Thanks!

PR #254, opened 2020-01-28
Add JSX API

Drup:
Follow up of #247 

cc @Khady @jorisgio @ulrikstrid


Drup:
Alright, the tests finally pass.

- The JSX doesn't depend on markup anymore. The common part of JSX and PPX are isolated in the tyxml-syntax package, and the parser-specific part is only in tyxml-ppx.
- The JSX supports namespaces ... reasonably. It's not perfect yet, but at least it tries to guess the correct one. On top, you can know specify a namespace by prefixing: `<Svg.Line> .... </Svg.Line>`. The first letter is turned lowercase (apparently, it's not syntactically valid to put a lowercase there ....).
- The list operator `<> ..... </>` is now recognized and supported
- Wrapping is supported

Current issues:
- All antiquotations must be single elements, unlike the PPX were they accept lists. List can only be spliced through the spread operator `...{foo}`. Unfortunately, the spread operator is super limited in where you can put it, which makes it impossible to write several things that the PPX would accept. For instance `[%html "<p>bar" elt1 "foo" elt2 "baz</p>"]` should translate to `<p> "bar" ...{elt1} "foo" ...{elt2} "baz" </p>` which isn't allowed
- The code transformation is currently global. I will add an attribute to turn it on/off, and I'm also considering accepting `<Tyxml.element />` to locally enable tyxml's jsx.

Drup:
I added the toggle. The syntax is `[@@@tyxml.jsx bool]` in ocaml, and `[@tyxml.jsx bool];` in reason.

@Khady I'm highly tempted to change the antiquotes so that in `<foo> "bar" other </foo>`, `other` must be a list of elements instead of a single one. A single element would be inserted by `([other])` (which is kinda verbose), and hopefully at some point, we can make reason accept `[other]` (without parens).
This would be much more consistent with the ppx, and more expressive.

Khady:
Hmm I’m probably not the best person to ask as I’m not using jsx extensively. But I tend to believe that I prefer the current setup.

- it is weird that “foo” is accepted and is a single element but others must be a list
- most of the time we don’t have a list of elements from what I gathered in our code
- when I write some HTML or use other template systems, each “element” is one element, not a list
- I feel that using ...others once in a while is less heavy than using [other] all the time

ulrikstrid:
I think keeping somewhat close to the react ppx would be preferable. I would rather have to spread a list than always wrap things with lists.

Drup:
My issue is that the spread operator is only available if it's the only child. You can't write `<> "foo" ...bar </>`.

Drup:
Everything's green, let's merge the current version! I kept the antiquotations as before. 

@Khady @ulrikstrid I'll let you test this for a week. If someone's motivated to write a reason-inclined documentation, I would appreciate it as well ^^'

PR #259, opened 2020-02-12
Preserve homemade elements.

Drup:
Homemade elements are distinguished by paths whose last part starts with an uppercase. Those will be preserved with the following spec:
- `<Foo a=x> children </Foo>` turns into `Foo.createElement ~a:x ~children:children ()`
- `<Foo a=x />` turns into `Foo.createElement ~a:x ()`
  No children at all.

Apparently, this is consistent with revery (cc @wokalski)

The list of children are interpreted according to the usual tyxml JSX rule and will be wrapped. In particular, if the tyxml is reactive, it'll be a reactive list.

@Khady, does that work for you ? :)
The previous version removed the `()` at the end, but keeping only labelled argument is a recipe for trouble, and @wokalski gave a fairly convincing case for removing the `~children` argument when the list is empty.

### DIFF
--- a/examples/basic_website_jsx/home_page.re
+++ b/examples/basic_website_jsx/home_page.re
@@ -1,0 +1,11 @@
+open Tyxml;
+
+let createElement = (~title: string, ~children: list('a), ()): Html.doc => {
+  <html>
+    <head>
+      <title> {Html.txt(title)} </title>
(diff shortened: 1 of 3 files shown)


Khady:
@Drup it seems to work with what we have. Thanks!

Drup:
Let's merge then!

PR #265, opened 2020-04-15
Improve the typing for restricted block constructs

Drup:
This strengthen a bit the typing for dt, address and header elements that
are not supposed to be nested.

In particular, this was allowed by tyxml (where it shouldn't):

```ocaml
let a = body [dl [dd [txt "foo"]; dt [a [h1 [txt "foo"]]]]]
```

More importantly (and that's the main reason for actually fixing this),
the coertion `flow5_without_sectioning_heading_header_footer :> flow5_without_header_footer`,
which was extremely annoying in practice.

### DIFF
--- a/lib/html_types.mli
+++ b/lib/html_types.mli
@@ -377,19 +377,30 @@ type labelable = [ | resetable | `Progress | `Meter | `Button ]
 
 type labelable_without_interactive = [ `Progress | `Meter]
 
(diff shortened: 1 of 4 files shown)


Drup:
I just realized this is not really backward compatible, which is mildly annoying to say the least. I'm wondering if anyone is affected except odoc.

PR #279, opened 2020-12-04
Update allowed inputmodes.

### LINKED ISSUE #278: inputmode="decimal"?
Hi there, I noticed tyxml rejects [`inputmode="decimal"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode). Should it be allowed? I also noticed several input modes I don't see referenced in the Mozilla docs (verbatim, latin, etc). I'm curious where the list comes from.

https://github.com/ocsigen/tyxml/blob/4d48e60269dc5d2f7ad62f8ef5af05b7b9e11042/lib/html_sigs.mli#L329-L333

joelburget:
Based on https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode.

Fixes #278.

### DIFF
--- a/lib/html_f.ml
+++ b/lib/html_f.ml
@@ -1021,6 +1021,8 @@ struct
     | `Tel -> "tel"
     | `Email -> "email"
     | `Url -> "url"
+    | `Text -> "text"
+    | `Decimal -> "decimal"
     | `Other s -> s
 
   let string_of_input_type = function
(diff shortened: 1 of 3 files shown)


lib/html_f.ml
```diff
@@ -1021,6 +1021,8 @@ struct
     | `Tel -> "tel"
     | `Email -> "email"
     | `Url -> "url"
+    | `Text -> "text"
+    | `Decimal -> "decimal"
```
joelburget, line 1025: Note: `Search` is already handled above.

Drup:
Thanks!

PR #296, opened 2022-01-12
Emit Unsafe.string_attrib for _-prefixed attributes in syntax ppxs to support non-standard attribute names

### LINKED ISSUE #295: Provide escape hatch for non-standard HTML attributes
Many Javascript libraries/frameworks depend upon HTML markup that includes non-standard attributes. Examples I've come across the past couple of days include:

* [AlpineJS](https://alpinejs.dev/)
* [htmx](https://htmx.org/)
* [unpoly](https://unpoly.com)
* [hyperscript](https://hyperscript.org/)

Of course, tyxml rejects these attributes, e.g. this hello-world example from htmx:

```html
<button hx-post="/clicked" hx-swap="outerHTML">
  Click Me
</button>
```

It is desirable to be able to use tyxml and benefit from the things it does well, while still being able to produce markup as required by libraries like these.

Conceptually, these custom attributes are basically just `data-*` attributes, and so there's no need for tyxml to do any kind of validation or checking on their values. All that is required is for tyxml to pass them along unharmed.

(The notion of configuring tyxml ppx and jsx-ppx (via dune ppx arguments) with a set of allowable additional attribute names was suggested in discord and discarded.)

cemerick:
Fixes #295 

(also includes more generalized kebab-casing for attributes from jsx)


cemerick:
FWIW, I've been using this in anger for a week or so, and it's working out quite nicely to emit [htmx](https://htmx.org/)-flavoured HTML.

Any thoughts on the approach?

Drup:
Sorry about the late response.

Interesting escape hatch. Why do you need `a_unchecked` instead of just `Unsafe.string_attrib` ?

Since `re` is already in the dependency cone, if your patch gets in, we definitely want to use that instead of `str`. There is also already some snake-to-caml case conversion tooling somewhere (in `/tools` maybe ?)

cemerick:
> Interesting escape hatch. Why do you need `a_unchecked` instead of just `Unsafe.string_attrib` ?

Oh, maybe I don't. I didn't think to (re)use `Unsafe` bits, but I guess that's what they're there for.

I'll tweak that up and swap `Str` -> `Re`.

cemerick:
I have updated my fork (and this PR) to:

1. Eliminate the use of `Str` in favor of `Re` (the case conversion stuff referenced `/tools` isn't sufficient; inputs to that `kebab_case` function can have a mix of kebab, snake, and camel case, alas)
2. Remove the `a_unchecked` and ``` `Unchecked ``` additions, using `Unsafe.string_attrib` instead. Hopefully my handling of the codegen in that case is reasonable (I didn't see any easy way to reuse any of the existing "parsers" there).

jsx/tyxml_jsx.ml
```diff
@@ -23,16 +23,22 @@ let lowercase_lead s =
...
-      first ^ "-" ^ lowercase_lead (String.sub name 4 (length - 4))
-    | _ -> name
-  else
-    name
+  let open Re in
+  let kebab string =
+    replace (Posix.compile_pat "[A-Z]") ~f:(fun g -> "-" ^ Group.get g 0) string
+    |> String.lowercase_ascii
+    |> replace_string (compile @@ char '_') ~by:"-" in
+  match exec_opt (Perl.compile_pat {|^(data_?|aria_?)(.+)|}) name with
```
Drup, line 31: Please put the compilation part at toplevel !

Drup (commented):
Nice work !

Could you adjust the documentation/tutorials for the ppx/jsx to mention this new feature ?

jsx/tyxml_jsx.ml
cemerick: Ah yeah I meant to do that before committing :upside_down_face:

cemerick:
I've updated to include top-level compilation of the necessary patterns, and added a section to the ppx/jsx docs.

(I haven't been able to "test" the latter though, as I had trouble getting the wikidoc to build; `make` in the `docs` directory does produce the API reference, but dies before it generates the narrative wiki content.)

Hopefully I didn't faff up the wiki documentation markup. :crossed_fingers:

Drup:
Thanks @cemerick

PR #298, opened 2022-02-10
[picture-tag] <img> should be the last child of a <picture> tag

YassineHaouzane:
According to :  
- [HTML Specs](https://html.spec.whatwg.org/multipage/embedded-content.html#the-picture-element)


Drup:
Indeed! Well caught. Merging as soon as CI is happy. :)

PR #299, opened 2022-02-10
[source-tag] we should allow srcset attribute in source tags

YassineHaouzane:
"The value of this attribute is ignored when the <source> element is
placed inside a \<picture\> element." According to : [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Source)

### DIFF
--- a/lib/html_types.mli
+++ b/lib/html_types.mli
@@ -1835,7 +1835,7 @@ type source_content = notag
 
 type source_content_fun = notag
 
-type source_attrib = [ | common | `Src | `Mime_type | `Media ]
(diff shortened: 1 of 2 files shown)


Drup:
So the attribute is allowed but ignored? If so, that change sounds good. Can you keep one of the tests with a normal source?

YassineHaouzane:
Yes I was trying to use the `picture` element but all my sources were ignored I fixed it by using the `srcset` attribute instead of the `src` one with the `Unsafe` module. The [HTML Specs](https://html.spec.whatwg.org/multipage/embedded-content.html#the-source-element) also says the same.

Drup:
Thanks!

Issue #288, opened 2021-07-13
Indentation option might add an extra space


Julow:
The following code renders differently in the browser whether the `indent` option is enabled or not:

```html
  <p>
   <span class="some padding ............................">some text here
    . and here
   </span>
  </p>
```

A line break is inserted between the two `txt` nodes. The extra breaks inside the `p` elements also change what `C-A, C-C` copies into the clipboard.

```ocaml
open Tyxml.Html

let content =
  html
    (head (title (txt "example")) [])
    (body
       [
         p
           [
             span
               ~a:[ a_class [ "some padding ............................" ] ]
               [ txt "some text here"; txt ". and here" ];
           ];
       ])

let () =
  Format.printf "%a" (pp ~indent:true ()) content
```

PR #300, opened 2022-02-10
Avoid breaking between text nodes

Julow:
I've attempted a fix to https://github.com/ocsigen/tyxml/issues/288 but it isn't working in every cases, for example this example is still problematic:

```ocaml
      p ~a:[ a_class [ "some padding ............................" ] ] [ 
        txt "some text here"; span [ txt ". and here" ]
      ]
```

There's an extra space before the `.` when indentation is enabled. Also happens in a more common example (in Odoc):

```html
  <p>This is a <a href="#">link</a>
  .
```

I don't think this can be solved on the printer side without adding metadata to node definitions, for example whether it is a "block" or "inline" element (defined for HTML only). Would this be reasonable ?

An other idea is to implicitly add `txt " "` nodes at the end of every block elements, which the printer could detect and decide how to break.

### DIFF
--- a/test/test_html.ml
+++ b/test/test_html.ml
@@ -89,7 +89,14 @@ let indent = "html indent", tyxml_tests ~indent:true Html.[
       txt "some text here"; txt ". and here"
     ]
   ],
-  "<p>\n <span class=\"some padding ............................\">\n  some text here. and here\n </span>\n</p>"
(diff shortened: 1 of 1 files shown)


Drup:
Sorry for the very late answer.

I don't like the solution of inserting `txt " "` at arbitrary points.

I think the solution is indeed to have a list of "elements where white space matters" and "elements where it does not", make the printer take those list as arguments, and define the list in the respective modules. What do you think about that ?

PR #301, opened 2022-03-05
Added dialog HTML element

sagotch:
Also added `a_onclose` attribute https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onclose


Drup (approved):
Thanks!

Issue #302, opened 2022-03-10
Extend `autocomplete` attribute values


aronerben:
Hello, thanks for this fantastic project!

Currently, the `autocomplete` attribute only supports `on` and `off`. The [spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete) lists many more options. It would be nice if `Tyxml` supported those too! I don't know if this is already on the roadmap, or similar, so I decided to open this issue. :) 

Encoding all possibilities mentioned in the spec with types seems a bit tedious (as there seem to be named groups, special cases for hidden input fields, etc.). Maybe simply turning it into a `string` attribute is "good enough" (I don't know what `Tyxml`'s philosophy is here). 

Currently we've solved it by unsafely adding the attribute like so:
```ocaml
let open Tyxml_xml in
let open Tyxml_html in
street_input
|> toelt
|> amap1 (fun _ attrs ->
      add_string_attrib "autocomplete" "street_address" attrs)
|> tot)
```
where `street_input` is some input field.

Drup:
That specification is indeed pretty awkward, but I would prefer to avoid "just using a string". In this case, it seems we could start with a fairly simple model such as ``` `On | `Off | `Tokens of string list ``` and see where that leads us. In any case the ppx extension can check more completely.

If you are willing, don't hesitate to submit a (potentially WIP) implementation ! There are instructions on how to add/change attributes in https://github.com/ocsigen/tyxml/blob/master/docs/Contributing.md . :)

aronerben:
@Drup Thanks for the quick response! I've drafted a PR: https://github.com/ocsigen/tyxml/pull/303

Drup:
Merged !

PR #303, opened 2022-03-10
Add token list option for autocomplete attribute on inputs

aronerben:
Implements a simplified version with enough leniency to represent the spec:
https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete

https://github.com/ocsigen/tyxml/issues/302


aronerben:
The tests are still missing and I added a question as a comment

lib/html_types.mli
```diff
@@ -2002,6 +2002,7 @@ type form_attrib =
     | `Method
     | `Name
     | `Target
+    (* TODO [aerben] forms only allow `On and `Off, how to deal with that? *)
```
Drup, line 2005: To answer your question: What's the behavior of web browser if you use the token list in a form ? Given what I know of html rendering, it's probably accepted silently, so we might as well allow it.

aronerben: I check in one of my own forms in Chromium (Version 102.0.5005.115) and if it's a token list it seems to default to `"on"`. Omitting the attribute also defaults to `"on"` in my browser (so I'm assuming it is, as you said, accepted silently).

aronerben:
> Thanks for the PR! I agree this change is for the best, so I would be inclined to merge with a few more test (and some changelog!) It's technically a breaking change, but it's a rather benign one. :)

Thanks for the review! Does the changelog suffice?

I force-pushed the branch to fix my authorship (I had forgot to set my username for this project). I added handling the case of `` `Tokens [] ``, which should default to `"on"` in my opinion.

PR #308, opened 2022-06-21
Fix `Svg.animate` type, name and attributes

rand00:


rand00:
I needed to change the type of Xml to include a `Semicolon` separator - which demands change to js_of_ocaml-tyxml too. I'll push a PR there too

Drup:
The change to the signature of `Xml.S` is a rather big breaking change (it breaks not only js_of_ocaml, but also eliom, and all users of the functorial API). 

I feel in this case, maybe it's for the best to first do all the renaming and deprecations, make a small release with all the other accumulated small fixes, then make a new big release (with some other breaking changes I accumulated).

rand00:
Okay - but one can't use the `Svg.animate` with animation-values and key-times without this major change; so makes `Svg.animate` quite useless anyway. Could postpone the renaming because of that?

rand00:
I was just hit by this bug of Tyxml again as I didn't use my fork in another project... TBH feels bad that a library focusing on typesafe HTML5 can't create correct SVG animations. Please either remove the `Svg.a_animation_values` and `Svg.a_keyTimes` or allow semicolon-separated lists :/

balat:
Thanks for finding this, and sorry it sat here for years. Picked up in #378.

You were right about the bug on both sides: commas on output, and the PPX
splitting those attributes on spaces, which makes `keySplines` impossible to
write since each control point group contains spaces.

I implemented it differently. Your patch adds a `Semicolon` constructor to the
`separator` type and `semicolon_sep_attrib` to `Xml_sigs.T`, which every
implementer of that signature has to follow -- js_of_ocaml-tyxml and eliom
included -- and I suspect that is why it stalled. #378 serialises in
`Svg_f.Wrapped_functions` instead, leaving `Xml_sigs` untouched at the cost of
one new function.

Keeping this open until #378 lands.

PR #309, opened 2022-08-12
ppx: textarea tag should be allowed to have no child 

ghuysmans:
The solution relies on previous work for empty `<script>` tags.

### DIFF
--- a/syntax/element_content.ml
+++ b/syntax/element_content.ml
@@ -218,7 +218,7 @@ let datalist ~lang ~loc ~name children =
   children::(nullary ~lang ~loc ~name [])
 
 
-let script ~lang ~loc ~name children =
+let at_most_one_child ~lang ~loc ~name children =
   match children with
   | [] ->
     let child = Common.txt ~loc ~lang "" in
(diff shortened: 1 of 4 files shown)


syntax/element_content.ml
```diff
@@ -228,6 +228,8 @@ let script ~lang ~loc ~name children =
     [Nolabel, child]
   | _ -> Common.error loc "%s can have at most one child" name
 
+let textarea = script
```
Drup, line 231: I would prefer you pulled the function as a `at_most_one_children` function, and alias that both time. :)

syntax/element_content.ml
```diff
@@ -228,6 +228,8 @@ let script ~lang ~loc ~name children =
...
+
```
Drup, line 232: A link to the spec in comment would be nice too.

Drup (approved):
Hi, thanks for the contribution. It looks good, I'll let you solve the small comments and merge! :)

syntax/element_content.ml
ghuysmans: Oh, right, that's way clearer.

syntax/element_content.ml
ghuysmans: Done; too bad I can't link the _Content model_ part.

Drup:
Oh, didn't realized you had merge rights. Thanks ! :)

ghuysmans:
Thank you for your review :)

PR #310, opened 2022-08-13
Introduce OCamlFormat

smorimoto:


smorimoto:
I found a bug in refmt during this journey, so I don't think it will be merged right away, so I would like to ask for your opinion on this in the meantime...

smorimoto:
The same solution is likely to help with tricky diff: https://github.com/llvm/llvm-project/blob/main/.git-blame-ignore-revs

Drup:
We should write a tool that find the ocamlformat options that minimizes the size of the diff, to fit a given project's indentation style :)

Personally, I use ocp-indent more often, with the options below. I don't use ocamlformat unless a project mandates it. Is the current setting the closest to that? In any case, I should probably play with ocamlformat to find what settings I like ...

```
normal
with=0
syntax=mll
max_indent=2
```

Drup:
I'm going to close this, as I don't think it necessary.

PR #312, opened 2022-09-28
fix build with OCaml 5 (drops support for 4.02)

rr0gi:
tyxml-ppx was already >= 4.04


aantron:
This would help users of Dream that want to use it with TyXML on OCaml 5.

Drup:
thanks !

PR #314, opened 2023-03-14
Typo in polymorphic variant `Gradient_Transform`.

### LINKED ISSUE #313: `GradientTransform` and `Gradient_Transform`
Hi,
Maybe I’m mixing two uses of `gradientTransform`, but I noted both usages of very the similar looking polymorphic variants `` `GradientTransform `` and `` `Gradient_Transform ``.
The type `lineargradient_attr` requires `` `GradientTransform ``, but the function `a_gradientTransform` produces `` `Gradient_Transform ``.
I think that they are meant to be the same: as-is, I can't use the function `a_gradientTransform` within a `linearGradient`.

Mbodin:
It should fix #313.

I'm surprised that I didn't had to change any implementation, but just the signature. Maybe we should add some tests.

### DIFF
--- a/lib/svg_sigs.mli
+++ b/lib/svg_sigs.mli
@@ -275,7 +275,7 @@ module type T = sig
     [< `UserSpaceOnUse | `ObjectBoundingBox ] wrap ->
     [ | `GradientUnits ] attrib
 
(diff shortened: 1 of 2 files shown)


Drup:
Thanks! Could you take this opportunity to add it as an svg test ?

PR #324, opened 2023-11-27
Typo whitout → without.

Mbodin:
This looks like a typo, so I made the corresponding changes.

Given that TyXML is used in several OCaml projects and that I'm changing the type specification, there might be other checks to do before accepting this PR as-is.


Drup:
Thanks ! I think I've seen that one, and didn't want to break compat for a typo :p 

Next version will be a major one, since there are other changes that are needed.
Can you add a change log entry ?

Mbodin:
Sure! I added one under the current version, but now that I think about it, should it be made apart if you are releasing a new major version?

PR #325, opened 2023-11-27
Adding support for `fill-opacity`

Mbodin:
I was in need of the `fill-opacity` attribute, so I did these changes. I haven't changed the parser yet, but I'm not sure where to start.

The attribute `fill-opacity` takes the same kind of argument than `offset` (a number or a percentage), but they mean different things: I think that it would not make sense to reuse the type `offset` (being either a number or a percentage), so I introduced a `number_or_percentage` type. Maybe we should instead declare several types for `offset` and `fill_opacity`?

### DIFF
--- a/lib/svg_f.ml
+++ b/lib/svg_f.ml
@@ -368,7 +368,7 @@ struct
   let a_fy = user_attrib string_of_coord "fy"
 
   let a_offset x =
-    user_attrib C.string_of_offset "offset" x
+    user_attrib C.string_of_number_or_percentage "offset" x
 
   let a_patternUnits x =
     user_attrib C.string_of_big_variant "patternUnits" x
@@ -538,6 +538,8 @@ struct
 
   let a_animation_fill x =
     user_attrib C.string_of_big_variant "fill" x
+
+  let a_fill_opacity = user_attrib C.string_of_number_or_percentage "fill-opacity"
 
   let a_fill_rule = user_attrib C.string_of_fill_rule "fill-rule"
 
@@ -1104,7 +1106,7 @@ struct
 
   let string_of_numbers_semicolon = list ~sep:"; " string_of_number
 
-  let string_of_offset = function
+  let string_of_number_or_percentage = function
     | `Number x -> string_of_number x
     | `Percentage x -> string_of_percentage x
 
--- a/lib/svg_sigs.mli
+++ b/lib/svg_sigs.mli
@@ -488,6 +488,10 @@ module type T = sig
   val a_animation_fill : [< | `Freeze | `Remove ] wrap -> [> | `Fill_Animation ] attrib
     [@@reflect.attribute "fill" ["animate"]]
 
+  val a_fill_opacity :
+    [< `Number of number | `Percentage of percentage ] wrap ->
+    [> | `Fill_opacity ] attrib
+
   val a_fill_rule : fill_rule wrap -> [> | `Fill_rule ] attrib
 
   val a_calcMode :
--- a/lib/svg_sigs.mli
+++ b/lib/svg_sigs.mli
@@ -1111,7 +1111,8 @@ module type Wrapped_functions = sig
 
   val string_of_numbers_semicolon : (float list, string) Xml.W.ft
 
-  val string_of_offset : ([< Svg_types.offset], string) Xml.W.ft
+  val string_of_number_or_percentage :
+    ([< `Number of number | `Percentage of percentage ], string) Xml.W.ft
 
   val string_of_orient : (Svg_types.Unit.angle option, string) Xml.W.ft
 
--- a/test/test_ppx.ml
+++ b/test/test_ppx.ml
@@ -416,6 +416,10 @@ let svg = "svg", SvgTests.make Svg.[
   [[%svg "<animate fill='freeze' values='1 2'/>"]],
   [animate ~a:[a_animation_fill `Freeze; a_animation_values ["1"; "2"]] []] ;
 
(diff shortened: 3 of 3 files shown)


Drup:
we do tend to try to use semantics names for the various possible inputs, but that's to be evaluated ton a case by case basis. Remind me what are the semantics in the offset and the opacity cases ?

Mbodin:
Right. Then it's better to define different types: I'll revert the renaming to offset and create a new type ☺

I think that offset relates to patterns: it moves the start of the pattern (for instance on dashed lines). If provided a percentage, then it's along the line for that percentage (and it can be negative or greater than 100%), if a number, it represents a distance (I don't know why one can't provide a unit there).

For fill-opacity, the percentage must be between 0% and 100% (it's just the opacity), and the number must be between 0 and 1. So it's not only used to represent different things (distance vs alpha), but also has different bounds.

I'm going to change this ☺

Drup:
Given your description ... maybe we could just mandate a float between 0 and 1 for the opacity, actually. The two representations are redundant.

Mbodin:
The attribute `fill-opacity` itself can be seen as redundant with `style='fill-opacity:__;'`, but it is useful as least for parsing. Is it fine to parse both percentages and numbers, but only using number internally?

Drup:
There are three things: 1) The API, 2) The representation 3) The parsing

Parsing should be complete (in this case, accepting both form), API should be convenient (In this case, avoid 15 layers of data format), and the internal representation should yield valid HTML5 (in this case, whatever works).

Mbodin:
Great! So let's write it such that we are able to write both percentages and numbers, but only storing numbers internally, and only printing out numbers ☺

I changed things in `reflect.ml` to be able to use my parser. I'm really not used to this kind of things (it feels like writing `Ltac` definitions in Coq). I hope that I did not do anything too wrong.

By the way, I noticed that there is a `fill_rule` parser defined in `syntax/attribute_value.ml` which is unused in `syntax/reflect/reflect.ml`. Is it normal?

Mbodin:
Sorry for the amount of commits: it took me some time to understand how everything works.

I'm starting to appreciate the point that I reached regarding the opacity attributes. What do you think?

syntax/attribute_value.ml
```diff
@@ -484,6 +484,33 @@ let paint ?separated_by:_ ?default:_ loc name s =
...
+          else n in
+
+        if v >= 0. && v <= 1. then
+          Some [%expr [%e (Common.float loc @@ v)]]
+        else
+          Common.error loc "Value of %s must be between 0 and 1." name
+
+      with Failure _ -> bad_form name loc
+
+    end [@metaloc loc]
```
Drup, line 512: That is a very spacious formatting :smile:

syntax/attribute_value.ml
```diff
@@ -484,6 +484,33 @@ let paint ?separated_by:_ ?default:_ loc name s =
...
+        let n = float_of_string (Re_str.matched_group 1 s) in
+
+        let v =
+          if group_matched 2 s then (n /. 100.)
+          else n in
+
+        if v >= 0. && v <= 1. then
+          Some [%expr [%e (Common.float loc @@ v)]]
+        else
+          Common.error loc "Value of %s must be between 0 and 1." name
```
Drup, line 508: This doesn't give a nice error message for inputs like `200%`.

lib/svg_types.mli
```diff
@@ -279,6 +279,8 @@ type strings = string list
 type color = string
 type icccolor = string
 
+type opacity = float
```
Drup, line 282: Can you add a comment that it's intended to be between O and 1 ? Same on the various `a_*_opacity` functions.

Drup (approved):
Looks excellent :+1: 

I made some misc comments, and we can merge just after.

Note that the use of `Re_str` is not mandatory, just tends to yield quick implementation.

syntax/attribute_value.ml
Mbodin: I tried to match the style of the rest of the file ☺ I agree it's quite spaced, and I'm fine with changing this.

syntax/attribute_value.ml
```diff
@@ -484,6 +484,28 @@ let paint ?separated_by:_ ?default:_ loc name s =
...
+      try
+        let n = float_of_string (Re_str.matched_group 1 s) in
+        let v =
+          if group_matched 2 s then (n /. 100.)
+          else n in
+        if v >= 0. && v <= 1. then
+          Some [%expr [%e (Common.float loc @@ v)]]
+        else
+          let (min, max) =
+            if group_matched 2 s then ("0", "1") else ("0%", "100%") in
```
Mbodin, line 504: I made a mistake here.

balat:
Thanks, and apologies for the long delay.

The opacity attributes are now on master through #372, which added the whole
set of missing SVG presentation attributes: `opacity`, `fill-opacity`,
`flood-opacity`, `lighting-color`, `stop-color` companions and the rest, all of
whose type tags were already declared in `presentation_attr` but had no
attribute function.

One deliberate difference from this PR: instead of introducing an
`alpha_value` type with its own `string_of_alpha_value`, the new functions
reuse `string_of_number`. Every addition to `Svg_sigs.Wrapped_functions` has
to be implemented again by everyone who implements that signature (eliom does,
by delegation), so #372 kept the number of new functions to exactly one, for
`a_download`. If a distinct `alpha_value` type turns out to be worth that
cost, it is easy to add on top.

Closing as covered. Thanks for the contribution!

PR #328, opened 2023-12-04
This should fix #326.

### LINKED ISSUE #326: Parser error for attribute transform
When I tried the following program:
```ocaml
let%svg test = "<g transform='translate(1,1)'></g>"
```
I got this error:
```
Error: Value of transform must be an SVG transform
```

I'm very surprised as the corresponding parser looks correct: https://github.com/ocsigen/tyxml/blob/d2916535536f2134bad7793a598ba5b7327cae41/syntax/attribute_value.ml#L350

Here follow the versions of my dependencies. I'm pinning to specific commits to get this fix: https://github.com/aantron/markup.ml/pull/69
```
tyxml-ppx: github:ocsigen/tyxml:tyxml-ppx.opam#d2916535536f2134bad7793a598ba5b7327cae41@a609c0aa
tyxml@opam:4.6.0@5ced2c2c
tyxml-syntax@opam:4.6.0@0e4fa721
markup: github:aantron/markup.ml#d77da228ed8915b6f11fbea00c269cfd5fd9e84a
```

My environment is a bit dirty at the moment. Does it reproduce on your side?
```[tasklist]
### Tasks
```

Mbodin:
I followed the specification at https://drafts.csswg.org/css-transforms/#svg-translate as much as I could. As such, I replaced white spaces by the more general `[ \t\r\f]`.

I tried to add some tests. I might not have placed them in the right place (I just noticed the message `This file is here to torture the ppx. Tests that are directly related to html or svg should go to the other files.` at the beginning of the file). Maybe some reordering should take place in the test file because of that :-\

What do you think?

### DIFF
--- a/syntax/attribute_value.ml
+++ b/syntax/attribute_value.ml
@@ -74,11 +74,11 @@ let list
   |> Common.list loc
   |> fun e -> Some e
 
-let spaces = list (Re_str.regexp " +") "space"
-let commas = list (Re_str.regexp " *, *") "comma"
-let semicolons = list (Re_str.regexp " *; *") "semicolon"
+let spaces = list (Re_str.regexp "[ \t\r\f]+") "space"
+let commas = list (Re_str.regexp "[ \t\r\f]*,[ \t\r\f]*") "comma"
+let semicolons = list (Re_str.regexp "[ \t\r\f]*;[ \t\r\f]*") "semicolon"
 
-let spaces_or_commas_regexp = Re_str.regexp "\\( *, *\\)\\| +"
+let spaces_or_commas_regexp = Re_str.regexp "\\([ \t\r\f]*,[ \t\r\f]*\\)\\|[ \t\r\f]+"
 let spaces_or_commas_ = exp_list spaces_or_commas_regexp "space- or comma"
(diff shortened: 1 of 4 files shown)


Mbodin:
I'm hitting another wall there: according to the spec, it seems that `skewY` expects a number (degrees are assumed here I think), but according to TyXML, it expects an angle: https://github.com/ocsigen/tyxml/blob/master/lib/svg_types.mli#L302 Maybe this changed between SVG 1.1 and SVG 2 ?

Mbodin:
I suggest to remove `skew` from the test and move it to another issue.

Mbodin:
Is it an issue not to support OCaml version 4.05 or before?

balat:
Thanks for this, and for the analysis in #326. Picked up in #376.

I took a different route for the `transform` parser itself: the reason
`translate(1,1)` failed is that `transforms` was reflected as
`spaces_or_commas transform`, so the value was split on commas *before* each
function was parsed, cutting `translate(1,1)` into `translate(1` and `1)`. #376
parses the list function by function instead, which removes the need for the
recursive regexp approach.

Your whitespace change is in, as `wsp` shared by the separator regexps, and your
test cases are in both the PPX and the JSX suites.

Keeping this open until #376 lands.

Issue #330, opened 2023-12-13
Spaces between tags are treated as pcdata


Mbodin:
Consider the following OCaml file:
```ocaml
open Tyxml
let%svg test = {|<g> </g>|} (* Notice the space between the opening and closing of the g tag. *)
```
The content of the g tag is interpreted as a ``[> `PCDATA ] Tyxml_svg.elt`` which is incompatible with `[< Svg_types.g_content ] Tyxml_svg.elt`, and I get a type error. In other words the space between the opening and closing of the g tag is interpreted as a pcdata. The same goes if I place a newline instead of the space.

I'm surprised as the SVG specification itself uses a lot of spacing in its examples. For instance the following SVG is provided as a valid example in https://www.w3.org/TR/2003/REC-SVG11-20030114/struct.html#GroupsOverview despite having the same spacing between its g tags.
```svg
<?xml version="1.0" standalone="no"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" 
  "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg width="4in" height="3in" version="1.1"
     xmlns="http://www.w3.org/2000/svg">
  <desc>Groups can nest
  </desc>
  <g>
     <g>
       <g>
       </g>
     </g>
   </g>
</svg>
```

I don't know the SVG specification enough to state what is meant to be there: should text with only white spaces be ignored instead of being treated as pcdata?

Drup:
We had a similar issue in HTML (I though it was in table). Some elements should have a special parsing mode to be white-space insensitive. We introduced it for appropriate HTML elements, so it's a matter of reusing that for the appropriate SVG elements.

Mbodin:
Thanks! I can see the discussion in #225.
I haven't find a place in the SVG specification listing which elements should be whitespace insensitive and which shouldn't. Would treating any tags not accepting pcdata as whitespace insensitive be too much?

Drup:
I'm not quite sure. @aantron , you probably have a bit more expertise there, do you have an opinion ?

aantron:
[...]

PR #331, opened 2023-12-15
Ignoring whitespaces in svg tags that do not accept pcdata.

Mbodin:
This PR aims at solving #330.


balat:
Thanks for this, and for the clear report in #330. Picked up in #377 with
tests.

Same approach as yours. I derived the list of elements from the content models
rather than writing it by hand: the 72 annotated elements are exactly those
whose `*_content` does not reach `` `PCDATA `` transitively, so `text`, `tspan`,
`textPath`, `desc`, `title`, `style`, `script` and `altGlyph` keep their
significant whitespace, and there is a test for that.

Keeping this open until #377 lands.

PR #335, opened 2024-03-04
[ci] Use the latest version of `actions/checkout`

ilankri:
To avoid the warning about deprecated Node.js 16 actions.  See https://github.com/ocsigen/tyxml/actions/runs/8134818335.


ilankri:
Superseded by #336.

PR #341, opened 2024-11-04
Add `name` and `ontoggle` attributes for `details` HTML elements.

SylvainBoilard:
Hello,

This PR adds support for the `name` attribute and specifying a `toggle` event handler for a `details` element.

See:
- https://html.spec.whatwg.org/multipage/interactive-elements.html#the-details-element ;
- https://html.spec.whatwg.org/multipage/indices.html#event-toggle ;
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#attributes ;
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#events .

Cheers,

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,10 @@
 * Add support for the clip-path presentation attribute
 	(#333 by Martin @MBodin Bodin)
 
+* Add support for the `name` attribute and `toggle` event on
+  `<details>` elements
+  (#341 by @SylvainBoilard)
+
 # 4.6.0
 
 * Update for OCaml 5.0 and drop support for OCaml 4.2.0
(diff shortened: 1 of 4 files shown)


Drup:
Thanks !

If I'm reading the spec right, `toggle` is a valid event for any popover element ... which can be any HTML element. In that case, `ontoggle` should be a global attribute.

Is that right ? Can you do the modification accordingly ?

SylvainBoilard:
Yes that is absolutely right; I had initially planned to limit the scope of this PR to the `details` element but following your input I decided to update it to support the `popover` API.

balat:
Hello @Drup what do you think of the new version?

balat:
Thank you, and sorry for the long wait.

Both halves of this pull request are now in master:

- The `popover` API (`a_popover`, `a_popovertarget`, `a_popovertargetaction`, and the `onbeforetoggle` and `ontoggle` handlers) landed in #366, as part of a wider pass over the WHATWG attributes.
- The `name` attribute on `details` landed in #382.

Closing as done. Thanks for the contribution.

PR #343, opened 2025-05-13
Add support for Microdata

toastal:
See: https://html.spec.whatwg.org/multipage/microdata.html

[^1]

[^1]: Please consider [giving up MS GitHub](https://sfconservancy.org/GiveUpGitHub/) or offering a non-proprietary, non-US-corporate-controlled mirror for this free software project. I wish to delete this Microsoft account in the future, but I need more projects like this to support alternative methods to send patches & contribute.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 * Add support for the clip-path presentation attribute
 	(#333 by Martin @MBodin Bodin)
+* Add support for Microdata attributes
+  (#343 by toastal)
+* Add support for <meta itemprop> to be in flow content
+  (#343 by toastal)
 * Fix typo for `subresource` & English orthographic fixes
   (#345 by toastal)
 


smorimoto:
Could you please rebase this?

balat:
Thank you for this, and sorry it sat for so long.

Everything this pull request adds is now in master, in two pieces:

- The five microdata global attributes (`a_itemid`, `a_itemprop`, `a_itemref`, `a_itemscope`, `a_itemtype`) landed in #366, as part of a wider pass over the WHATWG attributes.
- `meta` carrying an `itemprop` attribute as phrasing content landed in #382, with the `meta_itemprop` constructor and the `` `Meta_with_itemprop `` type tag you introduced here.

Two notes on the differences from this pull request:

- #382 adds only `` `Meta_with_itemprop `` to the phrasing content models, not the bare `` `Meta ``. The specification allows `meta` in a phrasing position only when it carries `itemprop`, and adding `` `Meta `` would also have allowed `<meta charset>` inside a `<p>`.
- `a_itemtype` first landed as a `string list`. You were right that the values are URLs: #385 makes it an `Xml.uri list`, using the `Xml.uri list` case you added to `reflect.ml`, and you are credited as co-author on that commit. The same reasoning applied to `a_ping`, which had the same problem.

Closing as done. Thanks again for the work and for the spec references.

PR #344, opened 2025-06-21
Allow `<div>` & script-supporting element in `<dl>`

toastal:
https://html.spec.whatwg.org/multipage/grouping-content.html#the-dl-element

Description list allow `<div>` & script-supporting elements—not just `<dd>` & `<dt>` elements.

[^1]

[^1]: Please consider [giving up MS GitHub](https://sfconservancy.org/GiveUpGitHub/) or offering a non-proprietary, non-US-corporate-controlled mirror for this free software project. I wish to delete this Microsoft account in the future, but I need more projects like this to support alternative methods to send patches & contribute.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@
 	(#333 by Martin @MBodin Bodin)
 * Fix typo for `subresource` & English orthographic fixes
   (#345 by toastal)
+* Allow `<div>` to wrap `<dd>` / `<dt>` as well as script-supporting elements
+  in `<dl>`
+  (#344 by toastal)
 
 # 4.6.0
 


balat:
Thanks, and apologies for the delay. Picked up in #375, rebased on master with
tests, quoting the two alternative content models of `dl` from the standard.

I kept your `script_supporting` category, and noted in the PR that the other
content models the standard allows script-supporting elements in (`ol`, `ul`,
`menu`, `table` and its sections, `tr`, `select`, `optgroup`) still do not
accept them, so that it does not get forgotten.

Keeping this open until #375 lands.

toastal:
would have been nice to see the commits that could apply be cherry-picked instead of swept into a single commit

balat:
You are right, and I am sorry. Your three commits should have been cherry-picked, with the rebase and the later adjustments as separate commits on top; instead the authorship was lost and only the prose credit and the CHANGES entry remain. The same happened to #328, #331, #348 and #308, so this is a process mistake on our side.

